PR #3, opened 2019-04-17
Float based viewbox

dirkschulze:


svgnative/src/SVGDocumentImpl.h
```diff
@@ -166,7 +167,7 @@ class SVGDocumentImpl
 
     XMLDocument mXMLDocument;
     // FIXME: Use floats internally.
```
qrb7xc: That FIXME is now fixed, right?

svgnative/src/SVGDocumentImpl.cpp
```diff
@@ -47,10 +47,7 @@ constexpr std::array<const char*, 5> gNonInheritedPropertyNames{{
 SVGDocumentImpl::SVGDocumentImpl(std::shared_ptr<SVGRenderer> renderer)
     : mRenderer{renderer}
 {
-    mViewBox.push_back(0);
-    mViewBox.push_back(0);
-    mViewBox.push_back(320);
-    mViewBox.push_back(200);
+    mViewBox = {{0, 0, 320.0f, 200.0f}};
```
qrb7xc: Add mViewBox to initialiser list above

svgnative/src/SVGDocumentImpl.cpp
```diff
@@ -69,33 +66,21 @@ void SVGDocumentImpl::TraverseSVGTree()
...
+            mViewBox[0] = SVGDocumentImpl::ParseLengthFromAttr(rootNode, "x", LengthType::kHorrizontal);
         if (HasAttr(rootNode, "y"))
-            bounds[1] = SVGDocumentImpl::ParseLengthFromAttr(rootNode, "y", LengthType::kVertical);
+            mViewBox[1] = SVGDocumentImpl::ParseLengthFromAttr(rootNode, "y", LengthType::kVertical);
         if (HasAttr(rootNode, "width"))
-            bounds[2] = SVGDocumentImpl::ParseLengthFromAttr(rootNode, "width", LengthType::kHorrizontal);
+            mViewBox[2] = SVGDocumentImpl::ParseLengthFromAttr(rootNode, "width", LengthType::kHorrizontal);
         if (HasAttr(rootNode, "height"))
-            bounds[3] = SVGDocumentImpl::ParseLengthFromAttr(rootNode, "height", LengthType::kVertical);
-        mViewBox.clear();
```
qrb7xc: It might be a bit unsafe to rely on the previous state of mViewBox if one of these attributes (x, y, width, height) is not present. But I see that TraverseSVGTree is called only once and right after the constructor so we could leave as is.

svgnative/src/SVGDocumentImpl.h
dirkschulze: can be removed now

dirkschulze:
The build issue seems to be related to a merge issue with git submodule.

PR #13, opened 2019-04-21
change protocol to access 3rd party repos, from ssh to https.

mpsuzuki:
currently, .gitmodule sets the protocol for 3rd party repos to ssh, like, "url = git@github.com:adobe/stylesheet.git". I suggest to change it to https, like, "url = https://github.com/adobe/stylesheet.git". It could be slightly more convenient to download & compile on the systems without registered ssh keys.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,7 +1,7 @@
 [submodule "third_party/stylesheet"]
 	path = third_party/stylesheet
-	url = git@github.com:adobe/stylesheet.git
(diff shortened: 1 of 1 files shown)


mpsuzuki:
Sorry, just I've signed with Adobe CLA, but the status seems to be kept. Should I close this PR and refile again?

mpsuzuki:
Thank you for taking care on this!

Issue #25, opened 2019-05-17
only DeprecatedSVGDocumentImpl.cpp uses "\r\n" ?


mpsuzuki:
If I'm not doing wrong manipulation, it looks like only DeprecatedSvgDocumentImpl.cpp uses MS-DOS-like line breaking, "\r\n". It might be expected that all sources use same manner (my preference is Unix-like "\n").

dirkschulze:
@mpsuzuki Only DeprecatedSvgDocumentImpl.cpp still does styling. The attributes are parsed with the SVGStringParser (currently). And that parser skips all kind of newlines `\r` and `\n` in any order.

Do you see any reason why we'd need to check for `\r\n` specifically anywhere but the (deprecated) CSS style parser? Or are we assuming that we would only get `\r\n` somewhere?

mpsuzuki:
Excuse me, maybe I could not explain what is my point due to my poor English. I have no objection about the design of the parser. I was just talking about the "\r" and "\n" in the source code.

For example, if I pass CSSColorKeywords.h to "od -c", the result includes only "\n", like this:

0000000   /   *  \n   C   o   p   y   r   i   g   h   t       2   0   1
0000020   9       A   d   o   b   e   .       A   l   l       r   i   g
0000040   h   t   s       r   e   s   e   r   v   e   d   .  \n   T   h

But, if I pass DeprecatedSvgDocumentImpl.cpp to "od -c", the result includes "\r\n", like this:

0000000   /   *  \r  \n   C   o   p   y   r   i   g   h   t       2   0
0000020   1   9       A   d   o   b   e   .       A   l   l       r   i
0000040   g   h   t   s       r   e   s   e   r   v   e   d   .  \r  \n

There is no technical problem. Just sometimes the editor complains. If there is a reason to use "\r\n" in this source, please ignore and close this issue. Sorry for spending your time!

dirkschulze:
@mpsuzuki I prefer unix's newline behavior as well. If you want to go through the files and correct the line ending, go ahead!

mpsuzuki:
Thank you for taking care on this trivial issue, I submitted a PR for this.

PR #28, opened 2019-05-21
replace CRLF (DOS-like newline) by LF (Unix-like newline) of the line breaking of the source

mpsuzuki:
Based on the discussion at the issue https://github.com/adobe/svg-native-viewer/issues/25 , I suggest to replace "\r\n" in DeprecatedSVGDocumentImpl.cpp by "\n".


mpsuzuki:
Thank you too for taking care on this!

Issue #18, opened 2019-04-30
Do not compile with style-element support by default


dirkschulze:
Right now, SVG Native Viewer gets compiled with some features that are deprecated.
* Add/remove custom style sheet
* Parse `<style>` element.
* Parse `style` attribute

This functionality is still used by some clients of SVG Native Viewer but shall be removed eventually. In the meantime, this code should be surrounded by preprocessor macros.

PR #30, opened 2019-06-02
Build w/ and w/o style support

dirkschulze:


dirkschulze:
Fixes #18 Compile w/o style support by default. Adapt tests.

PR #32, opened 2019-06-07
Create new ID to Element map

### LINKED ISSUE #26: Support rendering by 'id'
SVG OT allows adding multiple glyphs to the same SVG document. On rendering, the client needs to be able to choose which glyph to render by specifying an element node id.

dirkschulze:
Render only the element with a certain ID.

Still missing: Exposure to the API level. Fixes #26

### DIFF
--- a/svgnative/include/SVGDocument.h
+++ b/svgnative/include/SVGDocument.h
@@ -151,6 +151,11 @@ public:
      */
     void Render(const ColorMap& colorMap, float width, float height);
 
+    // TODO: Rather use an option based struct design?
+    void Render(const std::string& id, float width, float height);
+    void Render(const std::string& id, const ColorMap& colorMap);
+    void Render(const std::string& id, const ColorMap& colorMap, float width, float height);
+
 private:
     SVGDocument();
 
--- a/svgnative/include/SVGDocument.h
+++ b/svgnative/include/SVGDocument.h
@@ -151,9 +151,27 @@ public:
      */
     void Render(const ColorMap& colorMap, float width, float height);
 
-    // TODO: Rather use an option based struct design?
+    /**
+     * Renders the subtree of an element with the given XML ID.
+     */
+    void Render(const std::string& id);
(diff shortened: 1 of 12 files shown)


dirkschulze:
Still trying to add tests but would welcome early feedback to the API @mpsuzuki, @moazin and @qrb7xc.

svgnative/include/SVGDocument.h
```diff
@@ -151,6 +151,29 @@ class SVGDocument
      */
     void Render(const ColorMap& colorMap, float width, float height);
 
+    /**
+     * Renders the subtree of an element with the given XML ID.
+     */
+    void Render(const std::string& id);
```
dirkschulze: I'll replace `std::string` with a C string to avoid a copy for C wrappers.

svgnative/src/SVGDocumentImpl.h
```diff
@@ -202,14 +203,16 @@ class SVGDocumentImpl
 
     PropertySet ParsePresentationAttributes(XMLNode* node);
 
+    void RenderElement(const Element* element, const ColorMap& colorMap, float width, float height);
```
qrb7xc: Suggest to pass a const Element&

svgnative/src/SVGDocumentImpl.cpp
```diff
@@ -53,8 +53,8 @@ SVGDocumentImpl::SVGDocumentImpl(std::shared_ptr<SVGRenderer> renderer)
 
     GraphicStyleImpl graphicStyle{};
     std::set<std::string> classNames;
-    mGroup = std::unique_ptr<Group>(new Group(graphicStyle, classNames));
-    mGroupStack.push(mGroup.get());
+    mGroup = std::shared_ptr<Group>(new Group(graphicStyle, classNames));
```
qrb7xc: Suggest to use std::make_shared to create shared_ptrs

Issue #33, opened 2019-07-08
support of viewbox whose lowest-left point != (0,0)


mpsuzuki:
In SVG-OT fonts, often the graphics is drawn in the y<0 area. Thus, if we pass such kind of SVG to testSkia, the resulted PNG is slightly different from that by rsvg-convert etc.

For such cases whose lowest-left point != (0,0), client-side workarounds would be possible for Skia, Cairo, Qt backends, if we have the raw position of lowest-left point (viewbox.x and viewbox.y).

If I'm not misunderstanding, currently SVG Native Viewer does not provide an interface to retrieve viewbox.x and viewbox.y (we can retrieve viewbox.width and viewbox.height via SVGDocument::Width() and Height()). I made such functions at [get-offset-x-y branch](https://github.com/mpsuzuki/svg-native-viewer/commit/c88c5e95ea78c23131448d30ff4c7fdaf480261f)

I think some enhancement is required to make SVG Native Viewer works like rsvg, what is the right way?

mpsuzuki:
This is sample SVG, with viewbox (0,-50,100,100), whose lowest-left point is (0,-50).
[graphic-in-minus-y-area.zip](https://github.com/adobe/svg-native-viewer/files/3368678/graphic-in-minus-y-area.zip)

mpsuzuki:
This is a rasterization result by rsvg.
![graphic-in-minus-y-area rsvg](https://user-images.githubusercontent.com/2202675/60815962-9dda1f80-a1d3-11e9-8bdd-0ea9d8133f93.png)

mpsuzuki:
This is a rasterization result by testSkia. You can find the y<0 area is omitted, although resulted image is 100x100.
![graphic-in-minus-y-area skia](https://user-images.githubusercontent.com/2202675/60815998-b0545900-a1d3-11e9-9eeb-878253da97ef.png)

moazin:
#34 fixed this. :)

PR #34, opened 2019-07-13
Take into account viewbox translations too.

moazin:
At the moment, the library doesn't take care of any viewbox shifting as @mpsuzuki illustrates in #33. This is one way to fix that which works for me, though I am not sure whether it's the right way.


dirkschulze:
@moazin The change looks good to me. Would you be able to sign the Adobe CLA? Then I can merge the PR right away.

moazin:
> @moazin The change looks good to me. Would you be able to sign the Adobe CLA? Then I can merge the PR right away.

Sure. Just signed! :)

dirkschulze:
Great! Thank you very much for the PR!

PR #37, opened 2019-07-23
Fix the support for `use' elements.

### LINKED ISSUE #35: Support for `use` elements not working
While testing an [SVG document](https://github.com/adobe/svg-native-viewer/files/3395857/out.svg.txt) that I am attaching, I found some issues with the way `use` nodes are parsed and rendered. Due to these issues, the `use` element doesn't get rendered. The issues I noticed are as follows:

0. The value of `xlink:href` starts with a hash '#', however, `mResourceIDs` doesn't store the IDs with a hash. Thus the proper resource doesn't get found. `SVGDocumentImpl.cpp` lines [320-322](https://github.com/adobe/svg-native-viewer/blob/master/svgnative/src/SVGDocumentImpl.cpp#L320-L322)
```cpp
        auto resourceIt = mResourceIDs.find(href->value());
        if (resourceIt == mResourceIDs.end())
            return;
```
Not sure how this should be fixed, however for the purpose of this comment, I just removed the '#' from the SVG document (this isn't allowed, however, just for the rest of the points). 

1. In `SVGDocumentImpl.cpp`, line [330-332](https://github.com/adobe/svg-native-viewer/blob/master/svgnative/src/SVGDocumentImpl.cpp#L330-L332), are:
```cpp
        auto group = std::unique_ptr<Group>(new Group(graphicStyle, classNames));
        mGroupStack.push(group.get());
        AddChildToCurrentGroup(std::move(group));
```
This doesn't make sense to me. Firstly, a new group is being created. Then it's pushed on to the `mGroupStack`, then `AddChildToCurrentGroup` just adds the group itself in its `group->chilldren`. I think this should have been as:
```cpp
        auto group = std::unique_ptr<Group>(new Group(graphicStyle, classNames));
        auto tempGroupPtr = group.get();
        AddChildToCurrentGroup(std::move(group));
        mGroupStack.push(tempGroupPtr);
```
This will create a new group, add it to the `children` of the parent group and then push it in `mGroupStack`. 

2. Provided that I am correct about (1), even if I fix it, the SVG document I have attached doesn't render properly. The reason is on line [334](https://github.com/adobe/svg-native-viewer/blob/master/svgnative/src/SVGDocumentImpl.cpp#L334):
```cpp
        ParseChildren(resourceIt->second);
```
Here it is assumed that the `resourceIt->second` is a group-like element that has children. In the SVG document that I attach, that's not the case, it's just a `path` element, so `ParseChildren` just returns without doing anything. Thus nothing gets rendered. A quick but maybe bad fix would be:
```cpp
        if ( resourceIt->second->first_node() == 0 )
          ParseChild(resourceIt->second);
        else
          ParseChildren(resourceIt->second);
```

3. Even after fixing both (1) and (2), I don't get the expected result, the reason is that `use` elements can have styling applied, and that is not being taken care of in the `use` section. Once that is taken care of, the whole block will look like:
```cpp
        mFillStyleStack.push(fillStyle);
        mStrokeStyleStack.push(strokeStyle);
        auto group = [...] // code from (1) and (2) here
        mGroupStack.pop();
        mFillStyleStack.pop();
        mStrokeStyleStack.pop();
```
These fixes make it work on my end. I am not really sure if these are really bugs or if it's just my understanding of this library that is weak. In any case, let me know. If these are really bugs and should be solved, I can help by fixing these myself and submitting a PR. :-)

moazin:
This is an attempt to fix #35. The changes from line 320-323 are really crude. However, I am not too familiar with how `xlink:href` is supposed to behave.

### DIFF
--- a/svgnative/src/SVGDocumentImpl.cpp
+++ b/svgnative/src/SVGDocumentImpl.cpp
@@ -317,9 +317,15 @@ void SVGDocumentImpl::ParseChild(XMLNode* child)
         if (!href)
             return;
 
-        auto resourceIt = mResourceIDs.find(href->value());
+        if (href->value()[0] != '#')
+            return;
+
+        auto resourceIt = mResourceIDs.find((href->value() + 1));
         if (resourceIt == mResourceIDs.end())
             return;
+
+        mFillStyleStack.push(fillStyle);
+        mStrokeStyleStack.push(strokeStyle);
 
         auto transform = mRenderer->CreateTransform(
             1, 0, 0, 1, ParseLengthFromAttr(child, "x", LengthType::kHorrizontal), ParseLengthFromAttr(child, "y", LengthType::kVertical));
@@ -328,12 +334,18 @@ void SVGDocumentImpl::ParseChild(XMLNode* child)
         graphicStyle.transform = std::move(transform);
 
         auto group = std::unique_ptr<Group>(new Group(graphicStyle, classNames));
-        mGroupStack.push(group.get());
+        auto tempGroupPtr = group.get();
         AddChildToCurrentGroup(std::move(group));
+        mGroupStack.push(tempGroupPtr);
 
-        ParseChildren(resourceIt->second);
+        if(resourceIt->second->first_node() == 0)
+          ParseChild(resourceIt->second);
+        else
+          ParseChildren(resourceIt->second);
 
         mGroupStack.pop();
+        mFillStyleStack.pop();
+        mStrokeStyleStack.pop();
     }
     else if (elementName == "symbol")
     {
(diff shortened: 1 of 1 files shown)


svgnative/src/SVGDocumentImpl.cpp
```diff
@@ -317,23 +317,35 @@ void SVGDocumentImpl::ParseChild(XMLNode* child)
...
-        mGroupStack.push(group.get());
+        auto tempGroupPtr = group.get();
         AddChildToCurrentGroup(std::move(group));
+        mGroupStack.push(tempGroupPtr);
 
-        ParseChildren(resourceIt->second);
+        if(resourceIt->second->first_node() == 0)
+          ParseChild(resourceIt->second);
+        else
+          ParseChildren(resourceIt->second);
```
dirkschulze: Indentation seems off.

moazin: Sorry about that. Just changed that. Is it okay now?

dirkschulze:
Thanks a lot for the fix! Looks great.

PR #38, opened 2019-07-25
Support gradient elements outside `defs`.

### LINKED ISSUE #36: Support for `linearGradient` and related outside `defs` doesn't exist
In one of the test fonts we were given for OT-SVG that you can find [here](https://github.com/unicode-org/text-rendering-tests/blob/master/fonts/TestSVGmultiGlyphs.otf), I found an [SVG document](https://github.com/adobe/svg-native-viewer/files/3395980/out.svg.txt)
having a `radialGradient` element outside `defs` section. This is allowed in SVG specs, AFAIK. However, such an element is totally ignored by SVG Native Viewer.

moazin:
I am not sure if this would be the best way to fix it. Instead of creating multiple `else if` statements in `ParseChild` and copy-pasting sections of code from `ParseResource`, it's probably better to write one `else if` in `ParseChild` and let `ParseResource` do the actual job. 

This should close #36.

### DIFF
--- a/svgnative/src/SVGDocumentImpl.cpp
+++ b/svgnative/src/SVGDocumentImpl.cpp
@@ -370,8 +370,11 @@ void SVGDocumentImpl::ParseChild(XMLNode* child)
 
         mGroupStack.pop();
     }
-    else if (elementName == "style")
-        ParseStyle(child);
+    else if (elementName == "style" ||
+             elementName == "linearGradient" ||
+             elementName == "radialGradient" ||
+             elementName == "clipPath")
+        ParseResource(child);
 }
 
 void SVGDocumentImpl::ParseResources(XMLNode* node)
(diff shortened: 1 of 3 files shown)


dirkschulze:
@moazin Hi. I am really sorry for not looking at this earlier. Do you think you can put it to "Ready for review"?

moazin:
Sorry for not getting back on this earlier, I have been really busy. I think the following SVG file can work as a test, though I am not sure if the text output generated by the test suite is correct and should be used as the expected output. If you can verify that it is correct, I can commit these files.

[gradient-outside-defs.svg.txt](https://github.com/adobe/svg-native-viewer/files/3694597/gradient-outside-defs.svg.txt)
[gradient-outside-defs.txt](https://github.com/adobe/svg-native-viewer/files/3694598/gradient-outside-defs.txt)

dirkschulze:
@moazin I think the test looks fine. If you attach the diff to the PR as well, the test bots will verify the result. From what I can tell it looks reasonable.

dirkschulze:
@moazin Awesome! Looks like the tests passed! Thank you for the patch and taking the time to write a test.

PR #44, opened 2019-10-19
[WIP] Windraw port

dirkschulze:
This is a [Windraw port](https://github.com/mity/windrawlib). It allows using either GDI+ or D2D by putting a thin wrapper around both APIs.

We might not end up adding this wrapper but add a GDI+ and/or D2D ports individually.


dirkschulze:
Leaving the port up but canceling the PR.

Issue #51, opened 2019-10-31
Provide option to switch between static and dynamic library


dirkschulze:
Provide option to switch between static and dynamic library

dirkschulze:
Fixed

Issue #49, opened 2019-10-31
Choose and implement a version schema for svg-native-viewer


dirkschulze:
Use a version schema for the library. libtool's version schema preferred.

dirkschulze:
Fixed

PR #57, opened 2019-11-17
Add version for DLLs; option to create dynamic lib

dirkschulze:
Patch for issues #49 and #51.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,6 @@ A valid SVG Native document is always a valid SVG1.1/2.0 document.
 
 ### Known limitations in SVG Native Viewer
 * Referenced elements need to be declared first. Example: A `<linearGradient>` element must be defined in the SVG document before its first reference (`fill="url(#gradient)"`).
-* `viewBox` on `<svg>` element does not take translation values into account yet.
 * `preserveAspectRatio` is not supported on the `<svg>` element yet.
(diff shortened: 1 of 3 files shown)


dirkschulze:
@mpsuzuki Could I ask you to take a look at this patch please?

mpsuzuki:
@dirkschulze thanks, should I test this on Windows?

dirkschulze:
@mpsuzuki it is building on Windows. I was thinking of a review 🙂

mpsuzuki:
@dirkschulze OK, please let me spend some time to setup Windows development environment. The wordings VER_FILEVERSION, VER_PRODUCTVERSION are looking slightly odd for me (I think they are MS-specific terms). The versions of DLLs would be same for all platforms (Windows, macOS, and Unixs)?

mpsuzuki:
@dirkschulze Sorry delayed response. I've confirmed that proposed patch makes versioned DLL on Windows well. But during the linking of testSVGNative (the example for the text port), VS complains as they cannot find SVGNativeViewerLib.lib and aborts (please find a line before FAILED error). I would try to find some workarounds.
svg-native-viewer-vsbuild-error

dirkschulze:
@mpsuzuki you see this error because the examples link to SVGNativeViewerLib statically but there is only the dynamic version. You should see a similar problem on other OSs.

Issue #50, opened 2019-10-31
Provide .pc file for pkg-config


dirkschulze:
Provide .pc file for pkg-config

mpsuzuki:
@dirkschulze oh, sorry for leaving this issue as opened. I think this issue would be closed by PR #60

PR #59, opened 2019-11-20
Initial pkg-config file generation

dirkschulze:
#50

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 svgnative/build/
 viewer/
 .vscode
+SVGNativeViewerLib.pc
--- a/svgnative/CMakeLists.txt
+++ b/svgnative/CMakeLists.txt
@@ -293,3 +293,13 @@ if(NOT Boost_FOUND)
 else()
     target_include_directories(SVGNativeViewerLib PUBLIC ${Boost_INCLUDE_DIRS})
 endif()
+
+##############################
+# Creating pkg-config file
+##############################
+
+# TODO: Add includes and libraries for individual ports.
+set(BINARY_DIR "${CMAKE_CURRENT_BINARY_DIR}")
+set(SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}")
+set(PRIVATE_LIBS)
+CONFIGURE_FILE("${CMAKE_CURRENT_SOURCE_DIR}/SVGNativeViewerLib.pc.in" "${CMAKE_CURRENT_SOURCE_DIR}/SVGNativeViewerLib.pc" @ONLY)
--- a/svgnative/SVGNativeViewerLib.pc.in
+++ b/svgnative/SVGNativeViewerLib.pc.in
@@ -1,0 +1,10 @@
+prefix=@SOURCE_DIR@
+exec_prefix=@BINARY_DIR@
+libdir=${exec_prefix}
+includedir=${prefix}/include
+
+Name: SVGNativeViewerLib
+Description: Parser and renderer for SVG Native documents
+Version: @VERSION@
+
+Cflags: -I${includedir}


mpsuzuki:
@dirkschulze I guess current pc file is designed for a "built-but-not-installed" usecase (keeping sources, and the built libraries are staying at the working directory, but not installed). If my understanding is correct, naming "SVGNativeViewerLib-uninstalled.pc" is better to fit with the conventions. For example, cairo.pc is for the usecase which the built cairo libraries are installed (to /usr or /usr/local etc) and sources are already removed, and cairo-uninstalled.pc is for the usecase which the developer builds cairo and another library or another application dependent with cairo at the same time (both are in working directories and not installed). 

I should consider how "Libs" and "Libs.private" should be handled, but anyway, please let me know the assumed usecase of this. Should I ask to FreeType developers?

mpsuzuki:
oh, I found "make install" on GNU/Linux returns as

```
make: *** No rule to make target 'install'. Stop.
```

It's time to discuss the structure of the installed files, or suspend it and work as SVGNativeViewerLib-uninstalled.pc.

mpsuzuki:
@dirkschulze BTW, maybe current patch writes the configured pc file into the source directory. Maybe writing it into the working binary directory would be safer for the case like a developer builds for multiple Unix-like platforms under single source directory (as Build/linux and Build/mac). Please find  94a258b883d41aba04b4f98d81c8fedcf42bf3ca.

dirkschulze:
@mpsuzuki To be honest, I didn't work with pkg-config at all so far. I noticed that the binary path does not lead to the actual build target (Debug/ vs Release/ folder). One reason for that is that the final build path depends on the compiler system. On Linux, there wouldn't even be a Debug/ or Release/ folder. As a result, the script should work fine on Linux but maybe not on Windows.

The final destination of the binary can not be determined while running the configure because of the issue above. One way would be to add pre-processor macros for each system which add the build target folder structure. Not sure if that is the best way but it is the only way I can think of right now.

To un-install: I didn't even think about that nor did I look into `make install`. Do you have any experience and suggestions beside the final destination of the generated .pc file? (You may commit to the branch directly btw.)

mpsuzuki:
@dirkschulze I see, I understand why "install" target is not in current CMakeLists.txt.

The popular development style on GNU/Linux assumes as: the required libraries are installed as system-wide components, and pkg-config is the (very poor but) unified interface to query and retrieve the information how to use them. Of course, the users are expected to install the required libraries (as system-wide components) to develop or use the application. The modern package managers like apt (of Debian) or yum (of RedHat) would resolve the dependency and download/install the required libraries automatically. By these triggered installation, pc files are installed too, and they become discoverable.

But on MS Windows, such development style would be irregular. The applications are expected to be packaged with all required libraries, and the developer should not expect the users to install the third party libraries by themselves as system-wide components (it may cause DLL hell). Thus, in the popular application development for MS Windows, pkg-config (and pc files) is not useful so much. I guess, the developers on MS Windows would place the source or binaries of the required libraries into their building source tree (like svg-native-viewer/thirt_party). I guess the usage you assumed was something like such. In such case, without some tricky usage (to access uninstalled pc file), pkg-config command would be unable to find the pc files of the component libraries, it would not be helpful.

>I noticed that the binary path does not lead to the actual build target (Debug/ vs Release/ folder). 

Indeed. Because pkg-config was originally designed to be collaborate with GNU autotools which can build static & shared libraries at once, at the same directory, so its affinity with CMake is not good. Even if we can generate different pc files for Debug and Release, pkg-config command does not have the interface to choose one of them.

So... except of the developers working with Cygwin or Mingw, most developers on MS Windows would not be much careful about pkg-config file, I think.

>Do you have any experience and suggestions beside the final destination of the generated .pc file? (You may commit to the branch directly btw.)

Usually, the final destination of pc files (xxx.pc, designed to be used after the installation) would be /usr/lib/pkgconfig or /usr/local/lib/pkgconfig. The uninstalled pc files (xxx-uninstalled.pc, designed to be used after the building but before the installation) would be the top build directory (in the case of our README.md, Build/mac or Build/linux). Your question "how we can distinguish pc file for Release and for Debug?" is very reasonable, but pkg-config cannot serve for such request. So I think CMAKE_CURRENT_BINARY_DIR is acceptable.

Now I'm asking about the appropriate directory structures for the headers, to @lemzwerg and @moazin , please let me wait for their response, for a few days. I would commit something to pkg-config branch.

lemzwerg:
I suggest to put all header files into a subdirectory.  Assuming that the installation prefix is `/usr/local/`, this could be `/usr/local/include/svgnative`.

Another question is how DLL versioning will be done.  Dirk, what solution do you envision?  The standard solution is to decouple tar bundle versions from library versions, as described [here](https://www.gnu.org/software/libtool/manual/libtool.html#Versioning).

mpsuzuki:
@lemzwerg at present, DLL versioning is just done for Windows platform [PR#57](https://github.com/adobe/svg-native-viewer/pull/57) I guess, what you meant as "DLL versioning" was a general issue about the versioning of shared libraries, but calling it as DLL versioning might have misguided Dirk. As we have discussed about "make install" (and the assumed usage of this library) subsidy, the versioning of Win32 DLL is not so much important except of Cygwin or Mingw users. Soon I would tweak CMakeLists.txt for other platforms. As soon as I draft the tweak, I would open another Issue or PR, please let discuss in there.

mpsuzuki:
Just I've made my PR to pkg-config branch [60](https://github.com/adobe/svg-native-viewer/pull/60).

dirkschulze:
@mpsuzuki Can I close this PR in favor for #60 ?

mpsuzuki:
OK, let's continue discussion at there. Thanks!

PR #60, opened 2019-11-24
Add "make install" target, and to-be-installed & uninstalled pkg-config files

mpsuzuki:
This is a PR for "pkg-config" branch.

* Add "make install" target into CMakeLists.txt,
  to separate SVGNativeViewerLib.pc and SVGNativeViewerLib-uninstalled.pc.

* Use pkg-config to detect libjpeg availability.

Maybe there would be some discussion which is better name, "SVGNativeViewerLib" or "svgnative".


svgnative/CMakeLists.txt
```diff
@@ -297,9 +330,46 @@ endif()
...
+set(CMAKE_INSTALL_BINDIR "${CMAKE_INSTALL_BINDIR}")
+set(CMAKE_INSTALL_LIBDIR "${CMAKE_INSTALL_LIBDIR}")
+set(CMAKE_INSTALL_INCLUDEDIR "${CMAKE_INSTALL_INCLUDEDIR}")
+set(PRIVATE_LIBS "${PRIVATE_LIBS}")
+set(REQUIRES "${REQUIRES}")
+set(PRIVATE_REQUIRES "${PRIVATE_REQUIRES}")
+set(PORTS_INCLUDES "${PORTS_INCLUDES}")
+set(HEADER_SUBDIR "svgnative")
+CONFIGURE_FILE("${CMAKE_CURRENT_SOURCE_DIR}/SVGNativeViewerLib.pc.in"
+               "${CMAKE_CURRENT_BINARY_DIR}/SVGNativeViewerLib.pc" @ONLY)
```
lemzwerg: What about naming the file simply `svgnative.pc`?

svgnative/CMakeLists.txt
```diff
@@ -297,9 +330,46 @@ endif()
...
+CONFIGURE_FILE("${CMAKE_CURRENT_SOURCE_DIR}/SVGNativeViewerLib-uninstalled.pc.in"
+               "${CMAKE_CURRENT_BINARY_DIR}/SVGNativeViewerLib-uninstalled.pc" @ONLY)
```
lemzwerg: I've never seen a special 'uninstalled' version of a `.pc` file.  Is there another library that has something similar?

svgnative/CMakeLists.txt
mpsuzuki: The origin of the naming "SVGNativeViewerLib.pc" is the original patch by Dirk.
But this question is exactly what I wrote as "which is better name, "SVGNativeViewerLib or svgnative" :-). In my understanding, "SVG Native" is the name of a specification of restricted SVG. I cannot exclude a future possibility of other libraries related with SVG Native.

Myself has no special preference on this, I want to hear the comments by @dirkschulze.

svgnative/CMakeLists.txt
mpsuzuki: Please check the cairo source. There are [cairo.pc.in](https://cgit.freedesktop.org/cairo/tree/src/cairo.pc.in) and [cairo-uninstalled.pc.in](https://cgit.freedesktop.org/cairo/tree/src/cairo-uninstalled.pc.in). You can find uninstalled pc file writes the build directory pathname, but regular pc file does not have such.

svgnative/CMakeLists.txt
dirkschulze: I don't have a preference either. Can be the simpler name as well.

svgnative/CMakeLists.txt
```diff
@@ -64,10 +64,15 @@ CMAKE_DEPENDENT_OPTION(USE_CAIRO_EXAMPLE "Cairo example" TRUE "NOT LIB_ONLY;CAIR
 ################################
 if(CAIRO)
     find_package(PkgConfig)
-    include(FindCairo)
+    # include(FindCairo)
```
dirkschulze: Why is this commented out now? Can we remove it now? Does it require any system that wants to build with Cairo to use pkg-install? How is support on macOS?

svgnative/CMakeLists.txt
```diff
@@ -64,10 +64,15 @@ CMAKE_DEPENDENT_OPTION(USE_CAIRO_EXAMPLE "Cairo example" TRUE "NOT LIB_ONLY;CAIR
...
+    pkg_check_modules(CAIRO cairo)
     if(NOT CAIRO_FOUND)
-        set(CAIRO_INCLUDE_DIRS "/usr/include/cairo")
-        set(CAIRO_LIBRARIES "-lcairo")
+        message(FATAL_ERROR "Fatal error: Cairo port is requested, but CAIRO developers' files are missing.")
```
dirkschulze: So we reject building entirely now?

svgnative/CMakeLists.txt
dirkschulze: @lemzwerg concerns with this still?

svgnative/CMakeLists.txt
```diff
@@ -297,9 +330,46 @@ endif()
...
+CONFIGURE_FILE("${CMAKE_CURRENT_SOURCE_DIR}/SVGNativeViewerLib-uninstalled.pc.in"
+               "${CMAKE_CURRENT_BINARY_DIR}/SVGNativeViewerLib-uninstalled.pc" @ONLY)
+
+##############################
+# Installation
+##############################
+
+# library
+install(
+    TARGETS SVGNativeViewerLib
```
dirkschulze: Indentation seems inconsistent.

lemzwerg: No concerns.  I now understand why there are two `.pc` files.  However, I would still like to have simpler names for them :-)

dirkschulze:
@mpsuzuki Could you update the PR with the suggested names please?

mpsuzuki:
@dirkschulze Yes, of course. I found my comment-out of FindCairo.cmake was breaking the build for macOS (thank you for the mention on macOS), and small tweak for JPEG library is expected. I would update within 36 hours...

svgnative/CMakeLists.txt
mpsuzuki: @dirkschulze Yes, the old code was designed to try link cairo library even if it is not found by pkg-config. If a developer tries to build SVG Native Viewer with Cairo port, on the system without Cairo, the old code would cause an error. I think, reject building at the configuration phase might be easier to understand.

Or, should we disable Cairo port in the configuration phase (of course, some warning should be issued) is better? Other ports currently available (CoreGraphics, Skia, GDI+) have no check for the availability, so the discussion on Cairo availability would be the first discussion about such problem.

mpsuzuki: Oh, I overlooked how "-DSKIA=ON" is handled. Currently, -DSKIA=ON is disabled automatically if the platform is not macOS...

mpsuzuki: My current patch is still rejecting the building itself, if "-DCAIRO=ON" is given but cannot find working one. If you think "warn, disable and continue" is better, please comment.

svgnative/CMakeLists.txt
```diff
@@ -297,9 +348,46 @@ endif()
 ##############################
 # Creating pkg-config file
 ##############################
+include(GNUInstallDirs)
```
dirkschulze: Referring to the question about PkgConfig only being used with Cairo Graphics, shouldn't this be in a if clause for Cairo graphics as well?

svgnative/CMakeLists.txt
```diff
@@ -64,10 +64,23 @@ CMAKE_DEPENDENT_OPTION(USE_CAIRO_EXAMPLE "Cairo example" TRUE "NOT LIB_ONLY;CAIR
...
+       # FindJPEG.cmake does not use pkg-config, we do as FindCairo.cmake.
+       pkg_check_modules(PC_JPEG libjpeg)
+       if (PC_JPEG_FOUND)
+           FIND_PATH(JPEG_INCLUDE_DIRS NAMES jpeglib.h HINTS ${PC_JPEG_INCLUDEDIR})
+           FIND_LIBRARY(JPEG_LIBRARY NAMES jpeg HINTS ${PC_JPEG_LIBDIR})
+       else()
+           include(FindJPEG)
+       endif()
+       if (NOT PC_JPEG_FOUND AND NOT JPEG_FOUND)
+           message(FATAL_ERROR "Fatal error: Cairo port requires LIBJPEG developers' files.")
```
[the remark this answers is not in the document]
dirkschulze: Seems like there is no requirement later on.

mpsuzuki:
> Is there any special handling needed for -DSHARED=ON (dynamic library) in pkg?

For macOS and GNU/Linux, no special hook for shared library is required, I think. But for Win32, there would be a discussion how we control the symbols to export. At present, SVG Native Viewer source has no __declspec(dllexport) keywords to instruct MSVC whether the symbol should be exported or not. Thus, even if DLL file could be built, it has no symbols to be linked, so the export file (SVGNativeViewerLib.dll.lib) is created, and the resulted DLL is unusable.
In the changeset [50aa986f](https://github.com/mpsuzuki/svg-native-viewer/commit/50aa986f3f08372748e480335553e0aba94a2e36), I added CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS to expose all symbols by default.

svgnative/CMakeLists.txt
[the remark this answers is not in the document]
mpsuzuki: If Cairo port is built without JPEG decoder, Cairo port cannot rasterize the SVG including JPEG image. Other ports, like, Skia, CoreGraphics, GDI+, and Qt have JPEG decoder in their own, this is the reason why JPEG decoder is required for Cairo port.

svgnative/CMakeLists.txt
```diff
@@ -64,10 +64,23 @@ CMAKE_DEPENDENT_OPTION(USE_CAIRO_EXAMPLE "Cairo example" TRUE "NOT LIB_ONLY;CAIR
 ################################
 if(CAIRO)
     find_package(PkgConfig)
```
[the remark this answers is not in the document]
mpsuzuki: Oh, this "find_package(PkgConfig)" was introduced before the discussion about the requirement of pc files (it was introduced to check Cairo availability). I move it before "if(CAIRO)" at [e291f45b](https://github.com/mpsuzuki/svg-native-viewer/commit/e291f45b82b275224f48fc943da9c6ad45582495).

Anyway, generating .pc files do not require pkg-config command itself.

svgnative/CMakeLists.txt
mpsuzuki: Sorry for making you confused. Even if pkg-config command is unavailable, we can generate pc files. To obtain the infos to be written in pc files, "make install" target should be prepared, so, GNUInstallDirs.cmake module is useful, regardless with Cairo availability.

svgnative/CMakeLists.txt
```diff
@@ -121,6 +121,9 @@ if(WIN32)
   add_definitions(-DVER_PRODUCTVERSION=${VER_FILEVERSION})
   add_definitions(-DVER_PRODUCTVERSION_STR="${VERSION}")
 
+  if(USE_SHARED)
+    set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS YES CACHE BOOL "Export all symbols")
```
dirkschulze: I don't think we should do that, not even temporarily.

dirkschulze:
> > Is there any special handling needed for -DSHARED=ON (dynamic library) in pkg?
> 
> For macOS and GNU/Linux, no special hook for shared library is required, I think. But for Win32, there would be a discussion how we control the symbols to export. At present, SVG Native Viewer source has no __declspec(dllexport) keywords to instruct MSVC whether the symbol should be exported or not. Thus, even if DLL file could be built, it has no symbols to be linked, so the export file (SVGNativeViewerLib.dll.lib) is created, and the resulted DLL is unusable.
> In the changeset [50aa986f](https://github.com/mpsuzuki/svg-native-viewer/commit/50aa986f3f08372748e480335553e0aba94a2e36), I added CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS to expose all symbols by default.

I just checked with GCC and clang. Apparently we export all symbols there already. Not sure if something was not correctly setup in the CMakeList.txt file. We shouldn't do the same mistake on Windows and only export the symbols we want to be visible (in the public headers). IMO you shouldn't add this setting for now.

lemzwerg:
The 'exporting everything' option with gcc and clang is bad, however, it can be fixed rather easily; there are at least two solutions.

* Adding a proper export symbols definition file to the linker.
* Use the compiler's `-fvisibility=hidden` option and add `__attribute__((visibility ("default")))` to the functions that should be actually visible.  Maybe this can be wrapped into a macro so that it can simultaneously handle the `__declspec(dllexport)` case for Windows.

dirkschulze:
@lemzwerg I was mostly surprised that all symbols get exported even with clang. At least in Xcode this setting needs to get set explicitly. 

@mpsuzuki @lemzwerg I have created https://github.com/adobe/svg-native-viewer/pull/79 for the symbol export issue but was not able to verify it on Windows yet. Will try it on Windows ASAP.

mpsuzuki:
Just I've reverted the changeset 50aa986 making all symbol exposed on Windows platform.

dirkschulze:
@mpsuzuki IIRC you forked from another branch and didn’t create the PR against master. I don’t see the details on the mobile website. Could you take care of merging this PR? Thanks a lot!

mpsuzuki:
OK, can I merge this PR (as 0cf6c37) to master branch, by myself?

dirkschulze:
@mpsuzuki Yes of course. You should have enough privileges.

Issue #63, opened 2019-12-11
[Text Mode] NULL pointer dereference in SVGDocumentImpl::TraverseSVGTree


xinali:
Compile Options

```
ubuntu 16.04 x64
clang++-6.0 
ASAN
```

Here crash status

```
pwndbg> r poc_file_2
Starting program: /home/svg-native-viewer/svgnative/Build/linux/example/testText/test poc_file_2
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".

Program received signal SIGSEGV, Segmentation fault.
0x0000000000500d75 in boost::property_tree::detail::rapidxml::xml_base<char>::name (this=0x0) at /usr/include/boost/property_tree/detail/rapidxml.hpp:659
659                 return m_name ? m_name : nullstr();
LEGEND: STACK | HEAP | CODE | DATA | RWX | RODATA
────────────────────────────────────────────────────────[ REGISTERS ]────────────────────────────────────────────────────────
 RAX  0x0
 RBX  0x7fffffffd5e0 —▸ 0x7fffffffd980 —▸ 0x621000000100 ◂— 0x6cbb000000bfbbef
 RCX  0x0
 RDX  0x631000014800 ◂— 0x0
 RDI  0x0
 RSI  0x0
 R8   0x7fffffffd5b0 ◂— 0x0
 R9   0x7fffffffd5c0 —▸ 0x7fffffffd5a8 ◂— 0x0
 R10  0x7fffffffd4c0 ◂— 0x41b58ab3
 R11  0x10007fff7a98 —▸ 0xf1f1f1f1 ◂— 0x0
 R12  0x1
 R13  0x0
 R14  0x631000014800 ◂— 0x0
 R15  0x631000014800 ◂— 0x0
 RBP  0x7fffffffd4b0 —▸ 0x7fffffffd7f0 —▸ 0x7fffffffd950 —▸ 0x7fffffffe1a0 —▸ 0x5c2e50 (__libc_csu_init) ◂— ...
 RSP  0x7fffffffd490 ◂— 0x0
 RIP  0x500d75 ◂— cmp    qword ptr [rax], 0
─────────────────────────────────────────────────────────[ DISASM ]──────────────────────────────────────────────────────────
 ► 0x500d75    cmp    qword ptr [rax], 0
   0x500d79    je     0x500dad
    ↓
[...]

PR #71, opened 2019-12-20
Fix for Issue #63.

mpsuzuki:
* SVGDocumentImpl.cpp: Check whether the XMLNode is valid before using it.

### DIFF
--- a/svgnative/src/SVGDocumentImpl.cpp
+++ b/svgnative/src/SVGDocumentImpl.cpp
@@ -101,8 +101,7 @@ void SVGDocumentImpl::TraverseSVGTree()
 
 bool SVGDocumentImpl::HasAttr(XMLNode* node, const char* attrName)
 {
-    if (!node)
-        return false;
+    SVG_ASSERT(node != nullptr);
 
     auto attr = node->first_attribute(attrName);
     return attr != nullptr;
@@ -131,8 +130,7 @@ float SVGDocumentImpl::RelativeLength(LengthType lengthType) const
 
 float SVGDocumentImpl::ParseLengthFromAttr(XMLNode* node, const char* attrName, LengthType lengthType, float fallback)
 {
-    if (!node)
-        return fallback;
+    SVG_ASSERT(node != nullptr);
 
     auto attr = node->first_attribute(attrName);
     if (!attr)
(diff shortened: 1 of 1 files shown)


svgnative/src/SVGDocumentImpl.cpp
```diff
@@ -101,6 +101,9 @@ void SVGDocumentImpl::TraverseSVGTree()
 
 bool SVGDocumentImpl::HasAttr(XMLNode* node, const char* attrName)
 {
+    if (!node)
```
dirkschulze: Seems like this can be an assertion. We should check on the caller side.

svgnative/src/SVGDocumentImpl.cpp
```diff
@@ -128,6 +131,9 @@ float SVGDocumentImpl::RelativeLength(LengthType lengthType) const
 
 float SVGDocumentImpl::ParseLengthFromAttr(XMLNode* node, const char* attrName, LengthType lengthType, float fallback)
 {
+    if (!node)
```
dirkschulze: Ditto

svgnative/src/SVGDocumentImpl.cpp
```diff
@@ -389,6 +404,9 @@ void SVGDocumentImpl::ParseResources(XMLNode* node)
 
 void SVGDocumentImpl::ParseResource(XMLNode* child)
 {
+    if (!child)
```
dirkschulze: Can be an assertion too

svgnative/src/SVGDocumentImpl.cpp
```diff
@@ -835,6 +860,10 @@ void SVGDocumentImpl::ParseGraphicsProperties(GraphicStyleImpl& graphicStyle, co
 
 float SVGDocumentImpl::ParseColorStop(XMLNode* node, std::vector<ColorStopImpl>& colorStops, float lastOffset)
 {
+    /* XXX */
```
dirkschulze: What is this comment?

svgnative/src/SVGDocumentImpl.cpp
```diff
@@ -835,6 +860,10 @@ void SVGDocumentImpl::ParseGraphicsProperties(GraphicStyleImpl& graphicStyle, co
...
+    if (!node)
```
dirkschulze: Can be an assertion too

svgnative/src/SVGDocumentImpl.cpp
mpsuzuki: For the void functions, my patch returns immediately when invalid node is given.
But, ParseColorStop() should returns the floating number, so my patch returns 0, it would be hard for the client to find that: the node is broken and returned "0" is the fallbacked value.
OK, I would replace this by assertion.

Issue #64, opened 2019-12-12
[Text Mode] Infinite loop in CreatePath


xinali:
Compile option

```
ubuntu 16.04 x64
clang++-6.0 
ASAN
```

Infinite loop then ctrl+c, here is the status 

```
pwndbg> r poc_file_3
Starting program: /home/svg-native-viewer/svgnative/Build/linux/example/testText/test poc_file_3
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
^C
Program received signal SIGINT, Interrupt.
__memmove_avx_unaligned () at ../sysdeps/x86_64/multiarch/memcpy-avx-unaligned.S:157
157     ../sysdeps/x86_64/multiarch/memcpy-avx-unaligned.S: No such file or directory.
LEGEND: STACK | HEAP | CODE | DATA | RWX | RODATA
────────────────────────────────────────────────────────[ REGISTERS ]────────────────────────────────────────────────────────
 RAX  0x7fffdc800000 ◂— 0x0
 RBX  0x2
 RCX  0x5a20
 RDX  0x2
 RDI  0x7fffdc800000 ◂— 0x0
 RSI  0x5d5a20 ◂— and    byte ptr [rbp + 0x73], dh /* ' used)\n' */
 R8   0x10
 R9   0x24
 R10  0x83a150 (__asan::instance) —▸ 0x600000000000 ◂— 0x0
 R11  0x206
 R12  0x2
 R13  0x2
 R14  0x5db4e0 (.str) ◂— and    byte ptr [rdx], bl /* ' Z' */
 R15  0x7fffdc800000 ◂— 0x0
 RBP  0x7fffffffac60 ◂— 0x2
 RSP  0x7fffffffa3f8 —▸ 0x43652b (memcpy+1691) ◂— add    rsp, 0x838
 RIP  0x7ffff6c90b6f (__memmove_avx_unaligned+447) ◂— mov    word ptr [rdi], si
─────────────────────────────────────────────────────────[ DISASM ]──────────────────────────────────────────────────────────
 ► 0x7ffff6c90b6f <__memmove_avx_unaligned+447>    mov    word ptr [rdi], si
   0x7ffff6c90b72 <__memmove_avx_unaligned+450>    mov    word ptr [rdi + rdx - 2], cx
   0x7ffff6c90b77 <__memmove_avx_unaligned+455>    ret    
    ↓
[...]

PR #72, opened 2019-12-20
Fix for Issue #64.

mpsuzuki:
* SVGStringParser.cpp: increment the cursor 'pos' after taking 'Z' or 'z' operator, to prevent endless loop reported by Issue #64.

### DIFF
--- a/svgnative/src/SVGStringParser.cpp
+++ b/svgnative/src/SVGStringParser.cpp
@@ -404,6 +404,7 @@ void ParsePathString(const std::string& pathString, Path& p)
             return;
         if (!isDigit(*pos) && *pos != ',' && *pos != '-' && *pos != '.')
             prev = *pos++;
+        SVG_PARSE_TRACE("previous operator: " << prev);
         switch (prev)
         {
         case 'M':
@@ -455,6 +456,7 @@ void ParsePathString(const std::string& pathString, Path& p)
                 currentY = startY;
             }
 
+            prev = *pos++;
             break;
         case 'L':
             if (!ParseCoordinatePair(pos, end, currentX, currentY))


svgnative/src/SVGStringParser.cpp
```diff
@@ -455,6 +456,7 @@ void ParsePathString(const std::string& pathString, Path& p)
                 currentY = startY;
             }
 
+            prev = *pos++;
```
dirkschulze: I don't see the issue right now. We have `*pos++` at the beginning of the for loop if `pos` is not a number, nor -, + or a dot. It isn't since we have `Z`. So `pos` is looking at the next char already (which might be a serious of wsp, the next path command or invalid). Why advancing `pos` here once again?

`prev` should be `Z` here already which is correct. Btw. `prev` has the last path segment type.

dirkschulze: Also, could you add the test that runs into an infinite loop to the unit testing as well please?

mpsuzuki:
Sorry for making PR without understanding the implementation. I would send the poc files provided by @xina1i (in Issue #64), poc_file_3 is the file causing the infinite loop.

dirkschulze:
> Sorry for making PR without understanding the implementation. I would send the poc files provided by @xina1i (in Issue #64), poc_file_3 is the file causing the infinite loop.

Thanks a lot. Received them. No worries. The path syntax has some weird quirks which makes it hard sometimes. I think because of the investigation I even found one more issue. Will open it if it turns out to be one.

I created a PR with #75.

mpsuzuki:
I agree PR #75 is better solution, I withdraw this one.

Issue #65, opened 2019-12-13
[Text Mode] NULL pointer dereference in SVGNative::SVGDocument::Render


xinali:
Compile Options

```
ubuntu 16.04 x64
clang++-6.0 
ASAN
```

Debug

```
pwndbg> r poc_file_4                                                                                                         
Starting program: /home/tmp/svg-native-viewer/svgnative/Build/linux/example/testText/test poc_file_4

Program received signal SIGSEGV, Segmentation fault.
SVGNative::SVGDocument::Render (this=0x0) at /home/tmp/svg-native-viewer/svgnative/src/SVGDocument.cpp:59
59          mDocument->Render(colorMap, mDocument->mViewBox[2], mDocument->mViewBox[3]);
LEGEND: STACK | HEAP | CODE | DATA | RWX | RODATA
────────────────────────────────────────────────────────[ REGISTERS ]────────────────────────────────────────────────────────
 RAX  0x7fffffffc0f8 ◂— 0x0
 RBX  0x6d4c70 —▸ 0x4aee48 —▸ 0x407f10 ◂— lea    rsp, [rsp - 0x98]
 RCX  0x951bb0bf
 RDX  0x951b8a7c
 RDI  0x0
 RSI  0x5ad33ebe
 R8   0x6d5780 —▸ 0x6d400f ◂— 0x726f66736e617274 ('transfor')
 R9   0x1
 R10  0x126f
 R11  0x7ffff7ae1490 (__cxa_free_exception) ◂— mov    rax, qword ptr [rip + 0x2f1899]
 R12  0x7fffffffcd80 ◂— 0x9509119d
 R13  0x7fffffffc160 ◂— 0xe40
 R14  0x7fffffffc3c8 —▸ 0x6d4c70 —▸ 0x4aee48 —▸ 0x407f10 ◂— lea    rsp, [rsp - 0x98]
 R15  0x7fffffffc150 —▸ 0x6d3d00 ◂— '<svg xmlns'
 RBP  0x7fffffffc400 ◂— 'test-green'
 RSP  0x7fffffffc0f0 —▸ 0x6d4c80 —▸ 0x4b00c8 —▸ 0x41bf30 (SVGNative::StringSVGRenderer::~StringSVGRenderer()) ◂— lea    rsp, [rsp - 0x98]
 RIP  0x4092f2 (SVGNative::SVGDocument::Render()+98) ◂— mov    rdi, qword ptr [rdi]
[...]

PR #73, opened 2019-12-20
Fix for Issue #65.

mpsuzuki:
* SVGDocument.cpp: check the validity of mDocument before using it.

### DIFF
--- a/svgnative/src/SVGDocument.cpp
+++ b/svgnative/src/SVGDocument.cpp
@@ -55,51 +55,106 @@ SVGDocument::~SVGDocument() {}
 
 void SVGDocument::Render()
 {
+    if (!mDocument)
+        return;
+
     ColorMap colorMap;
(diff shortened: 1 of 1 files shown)


svgnative/src/SVGDocument.cpp
```diff
@@ -55,51 +55,106 @@ SVGDocument::~SVGDocument() {}
...
+{
+    if (!mDocument)
+        return;
+
+    mDocument->Render(id, colorMap, width, height);
+}
+
+std::int32_t SVGDocument::Width() const
+{
+    return static_cast<std::int32_t>(mDocument->mViewBox[2]);
```
dirkschulze: Isn't the mDocument check missing here too?

mpsuzuki: Indeed, it was my mistake. Sorry, I added it.

Issue #69, opened 2019-12-19
rx == 0 or ry == 0 ellipse or rectangle can paralyze Cairo port, at least.


mpsuzuki:
During the investigation of my patch for Issue #67, I found my cairo port is paralyzed by the ellipse or rectangles with rx == 0 or ry == 0. I think Skia and CoreGraphics ports are safe from this issue, because they have direct functions to draw ellipse or rounded-rectangle.

Cairo has no direct functions to draw ellipse or rounded-rectangle, so I inserted cairo_scale() to stretch the circle or rounded-corner of the rounded rectangle. Maybe, cairo_scale(cr, 0, scale_y) or cairo_scale(cr, scale_x, 0) and following cairo_arc() makes cairo confused, and all following objects are not drawn correctly.

So, I drafted a patch to replace ["rx == 0 or ry == 0 ellipse" by a squashed bounding-box](741849aba4f98fbd5d7c3fbceeaa46a21f9ff331), and to replace ["rx == 0 or ry == 0 rounded rectangle" by normal rectangle](0df194c5a36d5e694996858db87feaab12bac286).

Another option might be "if an ellipse or a rounded-rectangle with rx==0 or ry==0, do not take it as a drawable element". I wish if the experts give some comments on the direction we should go.

mpsuzuki:
If rx and ry are non-zero, like,
```
<svg xmlns="http://www.w3.org/2000/svg" width="300" height="200" viewBox="0 0 300 200">
  <ellipse cx="150" cy="50" rx="25" ry="25"/>
  <rect x="120" y="120" width="60" height="60" rx="15"  ry="15"/>

  <ellipse cx="50"  cy="50" rx="1" ry="25"/>
  <rect x="20"  y="120" width="60" height="60" rx="1"   ry="15"/>

  <ellipse cx="250" cy="50" rx="50" ry="25"/>
  <rect x="220" y="120" width="60" height="60" rx="150" ry="15"/>
</svg>
```

The result is looking like
![rect2-fixed](https://user-images.githubusercontent.com/2202675/71163083-bb85d780-228f-11ea-8b1d-e597cf003a02.png)

mpsuzuki:
But, if rx == 0, like,
```
<svg xmlns="http://www.w3.org/2000/svg" width="300" height="200" viewBox="0 0 300 200">
  <ellipse cx="150" cy="50" rx="25" ry="25"/>
[...]

Issue #67, opened 2019-12-14
[text mode]ParseShape(SVGNative::XMLNode*): Assertion `rx == ry' failed.


hardik05:
found one issue where program terminates because of assert failure, following is the log:
pwndbg> r /home/test/svg-native-viewer/svgnative/Build/linux/example/testText/output/Slave/crashes/poc.svg 1
Starting program: /home/test/svg-native-viewer/svgnative/Build/linux/example/testText/testSVGNative /home/test/svg-native-viewer/svgnative/Build/linux/example/testText/output/Slave/crashes/poc.svg 1
testSVGNative: /home/test/svg-native-viewer/svgnative/src/SVGDocumentImpl.cpp:511: std::unique_ptr<SVGNative::Path> SVGNative::SVGDocumentImpl::ParseShape(SVGNative::XMLNode*): Assertion `rx == ry' failed.

Program received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
51      ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
LEGEND: STACK | HEAP | CODE | DATA | RWX | RODATA
───────────────────────────────────────────────────────────[ REGISTERS ]───────────────────────────────────────────────────────────
 RAX  0x0
 RBX  0x0
 RCX  0x7ffff70e3e97 (raise+199) ◂— mov    rcx, qword ptr [rsp + 0x108]
 RDX  0x0
 RDI  0x2
 RSI  0x7fffffffd460 ◂— 0x0
 R8   0x0
 R9   0x7fffffffd460 ◂— 0x0
 R10  0x8
 R11  0x246
 R12  0x555555619310 ◂— '/home/test/svg-native-viewer/svgnative/src/SVGDocumentImpl.cpp'
 R13  0x5555556193b3 ◂— jb     0x55555561942d /* 'rx == ry' */
 R14  0x1ff
 R15  0x0
 RBP  0x7ffff725c7d8 ◂— and    eax, 0x25732573 /* "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n" */
 RSP  0x7fffffffd460 ◂— 0x0
 RIP  0x7ffff70e3e97 (raise+199) ◂— mov    rcx, qword ptr [rsp + 0x108]
────────────────────────────────────────────────────────────[ DISASM ]─────────────────────────────────────────────────────────────
[...]

PR #74, opened 2019-12-21
Disable rendering when we are close to 0 for dimension attributes on …

dirkschulze:
…circle, ellipse and rect.

I am using epsilon to determine if we are close to 0. We could support smaller values around 0 but anything smaller than that is less likely to occur in the real world but more likely to cause issues in any of the rendering libraries.

Fixes issues: #69, #67

### DIFF
--- a/svgnative/src/SVGDocumentImpl.cpp
+++ b/svgnative/src/SVGDocumentImpl.cpp
@@ -16,6 +16,7 @@ governing permissions and limitations under the License.
 #include "SVGStringParser.h"
 
 #include <cmath>
+#include <limits>
 
 namespace SVGNative
 {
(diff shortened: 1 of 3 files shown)


mpsuzuki:
Do not draw a rectangle or an ellipse looking like a line, is good idea. However, yet I'm unsure which is better: draw nothing, or draw a line. I remember, a rectangle operation is often used to draw a line. My criteria would be:
- if future revision of "SVG Native" spec mentions the cases which the renderers may ignore in the rendering, the patch is really OK.
- even if future revision of "SVG Native" spec does not mention this, if the stroke-width is taken into account too, ignoring the objects with too small rx, ry and stroke-width are OK.
- for recognizable stroke-width objects, replacing by the line might be safer.

How do you think about?

dirkschulze:
@mpsuzuki The spec SVG Native builds upon is the SVG 2 spec which is clear about this: https://svgwg.org/svg2-draft/shapes.html#EllipseElement

mpsuzuki:
@dirkschulze Thanks, now I understand the proposed patch is exactly matching with the spec.

PR #79, opened 2019-12-24
Export SVGDocument and renderer symbols only

dirkschulze:


mpsuzuki:
I compared the list of exported symbols (DUMPBIN.EXE /EXPORTS output) for my "export all" patched and this "visibility patched", and confirmed the number of exported symbols are really slimmed, from 735 to 50. Of course, I confirmed testSVGNative.exe, testSkia.exe could be linked (and executable) with the DLL after visibility patch.
[exported-sym-list.zip](https://github.com/adobe/svg-native-viewer/files/3999815/exported-sym-list.zip)

PR #83, opened 2019-12-26
CoreGraphics port: add "-framework CoreFoundation" to the linker flag to use libSVGNativeViewerLib.

mpsuzuki:
d1f8a2c cannot finish the linking of testC. To finish it, "-framework CoreFoundation" is added to the linker flag in CMakeLists.txt.
An archive library does no record its dependency, so the dependencies should be supplied by the linker flags explicitly, also it is added to pkg-config file.


dirkschulze:
@mpsuzuki Thanks! Was planning to do it today. Merging.

mpsuzuki:
Thanks! I'm happy to see Green badge on this branch!!

Issue #82, opened 2019-12-25
operation to draw rounded rectangle for rx != ry cases


mpsuzuki:
At present, Path->RoundedRect() method is for the case rx == ry. SVG spec allows to users to have a rounded-rect with different rx and ry. The codes in each ports are expected to be improved to support rx != ry cases.

I added another RoundedRect() taking x, y, width, height, rx and ry (existing one takes x, y, width, height and r) to SVGRender.h, and its real implementations to CoreGraphics, Skia and Cairo. I already drafted for GDI+ and Qt, but not tested yet.

mpsuzuki:
My patch working in progress is here: https://github.com/mpsuzuki/svg-native-viewer/compare/PR-enhance-rounded-rect

CoreGraphics port is confirmed.
![image](https://user-images.githubusercontent.com/2202675/71439547-6d644e80-273d-11ea-938a-87ebd86d8018.png)

Skia port is confirmed (on macOSX).
![rrect-mac-skia](https://user-images.githubusercontent.com/2202675/71439682-e82d6980-273d-11ea-980c-5d57502da9f1.png)
![rrect-mac-skia2](https://user-images.githubusercontent.com/2202675/71439708-05623800-273e-11ea-93b1-48cebf6fb3ed.png)

Soon I would upload the confirmation by Cairo.

mpsuzuki:
Here are Cairo results (upper is not patched, lower is patched)

![Cairo result, current](https://user-images.githubusercontent.com/2202675/71446133-651f0a00-2763-11ea-9ffe-6c9681defd20.png)
![Cairo result, patched](https://user-images.githubusercontent.com/2202675/71446142-75cf8000-2763-11ea-8d56-6d167109c134.png)

mpsuzuki:
Here are GDI+ results (upper is not patched, lower is patched)
![rrect gdiplus-current](https://user-images.githubusercontent.com/2202675/71460352-d8219280-27ee-11ea-8337-6f165a4799e7.png)

![rrect gdiplus-fixed](https://user-images.githubusercontent.com/2202675/71460377-eec7e980-27ee-11ea-9dcf-b06ef4f528a8.png)

mpsuzuki:
I tried to modify testGDIPlus to take an argument to specify a SVG file to be drawn, but my Win32 programming experience is too poor. I replaced hardwired SVG in testGDIPlus.

mpsuzuki:
Because I submitted PR #84 for this issue, I close here.

PR #84, opened 2019-12-26
Add RoundedRect() for rx != ry cases.

mpsuzuki:
* Designed for the code after 'epsilon' check @ 45fc8f49
* CoreGraphics, GDI+ and Cairo port results are shown at Issue #82.

### DIFF
--- a/svgnative/include/SVGRenderer.h
+++ b/svgnative/include/SVGRenderer.h
@@ -218,6 +218,7 @@ public:
 
     virtual void Rect(float x, float y, float width, float height) = 0;
     virtual void RoundedRect(float x, float y, float width, float height, float cornerRadius) = 0;
+    virtual void RoundedRect(float x, float y, float width, float height, float cornerRadiusX, float cornerRadiusY) = 0;
     virtual void Ellipse(float cx, float cy, float rx, float ry) = 0;
 
     virtual void MoveTo(float x, float y) = 0;
--- a/svgnative/include/SVGRenderer.h
+++ b/svgnative/include/SVGRenderer.h
@@ -217,7 +217,6 @@ public:
     virtual ~Path() = default;
 
     virtual void Rect(float x, float y, float width, float height) = 0;
-    virtual void RoundedRect(float x, float y, float width, float height, float cornerRadius) = 0;
     virtual void RoundedRect(float x, float y, float width, float height, float cornerRadiusX, float cornerRadiusY) = 0;
     virtual void Ellipse(float cx, float cy, float rx, float ry) = 0;
 
--- a/svgnative/src/SVGDocumentImpl.cpp
+++ b/svgnative/src/SVGDocumentImpl.cpp
@@ -527,7 +527,7 @@ std::unique_ptr<Path> SVGDocumentImpl::ParseShape(XMLNode* child)
         ry = std::min(ry, height / 2.0f);
 
         auto path = mRenderer->CreatePath();
-        if (rx == 0 || ry == 0)
+        if (isCloseToZero(rx) || isCloseToZero(ry))
         {
             path->Rect(x, y, width, height);
         }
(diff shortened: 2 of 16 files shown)


svgnative/include/SVGRenderer.h
```diff
@@ -218,6 +218,7 @@ class Path
 
     virtual void Rect(float x, float y, float width, float height) = 0;
     virtual void RoundedRect(float x, float y, float width, float height, float cornerRadius) = 0;
+    virtual void RoundedRect(float x, float y, float width, float height, float cornerRadiusX, float cornerRadiusY) = 0;
```
dirkschulze: If we are changing (by adding new functions to) the SVGRenderer, why not just have one RoundedRect function. Deriving classes can optimize/differ between rx == ry and rx != ry.

dirkschulze: Also, please add tests to make sure we execute all the cases of (non-)rounded rects.

mpsuzuki: So, should I replace existing RoundedRect(x, y, w, h, r) by RoundedRect(x, y, w, h, rx, ry)?

dirkschulze: yes. I think so.

svgnative/test/elem-currentColor.txt
```diff
@@ -1,5 +1,9 @@
 [group transform: matrix(1,0,0,1,0,0)
-    [path Rect(0,0,200,200)
-        fill: {hasFill: true winding: nonzero paint: rgba(1,0,0,1)}
-        stroke: {hasStroke: false width: 1 cap: butt join: miter miter: 4 dashOffset: 0 paint: rgba(0,0,0,1)}]
+    [group
+        [group
```
dirkschulze: Why the additional groups?

svgnative/test/elem-defs.txt
```diff
@@ -0,0 +1,4 @@
+[group transform: matrix(1,0,0,1,0,0)
```
dirkschulze: Where is the test for this? Just see the result.

svgnative/test/elem-transform-on-parent.txt
```diff
@@ -1,5 +1,9 @@
 [group transform: matrix(1,0,0,1,0,0)
-    [path Rect(0,0,200,200)
-        fill: {hasFill: true winding: nonzero paint: rgba(0,0.502,0,1)}
-        stroke: {hasStroke: false width: 1 cap: butt join: miter miter: 4 dashOffset: 0 paint: rgba(0,0,0,1)}]
+    [group
+        [group transform: matrix(0.707,0.707,-0.707,0.707,0,0)
```
dirkschulze: Why the 2 new groups and the additional transform?

dirkschulze:
I am not sure why but CircleCI didn't pick up this PR it seems. Need to check.

dirkschulze:
@mpsuzuki I still don't know why CircleCI doesn't test your most recent PRs. I just put up another PR myself and the bots do run. I checked and CircleCI is setup to build PRs from forked repos. IIRC the bots did run for you before, didn't they?

mpsuzuki:
I'm quite sorry, I made a mistake to regenerate text data. I should use something like "runTest.py --output=regenerated-texts", but I executed testSVGNative manually (I found runTest.py has a special hook for the test files "elem-*.svg").
By using runTest.py, the results are same with before, except of the SVG including rounded rect, clipping.svg. I reverted d866b60.

Sorry for my delay to check why the CircleCI does not execute CI checks. I would check from now.

svgnative/src/SVGDocumentImpl.cpp
```diff
@@ -527,14 +527,13 @@ std::unique_ptr<Path> SVGDocumentImpl::ParseShape(XMLNode* child)
         ry = std::min(ry, height / 2.0f);
 
         auto path = mRenderer->CreatePath();
-        // TODO: Create path for rectangle if rx != ry.
-        if (rx == 0 && ry == 0)
+        if (rx == 0 || ry == 0)
```
dirkschulze: Last note, we should use `isCloseToZero(rx) || isCloseToZero(ry)` here since it is rendering related.

PR #86, opened 2019-12-26
replace example/testC/TestMain.cpp by TestMain.c

mpsuzuki:
For the developers' reference, C program is better reference to explain C-wrapper API.

* SVGNativeCWrapper.{cpp,h}: svg_native_render_output() API tweaked.
* example/testC/TestMain.c: rewrite of TestMain.cpp.
* example/testC/TestMain.cpp: removed.
* example/testC/CMakeLists.txt: TestMain.cpp is replaced by TestMain.c.


svgnative/src/SVGNativeCWrapper.cpp
```diff
@@ -239,20 +239,25 @@ void svg_native_render_size(svg_native_t* sn, float width, float height)
 }
 
 #ifdef USE_TEXT
-void svg_native_render_output(svg_native_t* sn, char* output, int& length)
+size_t svg_native_render_output(svg_native_t* sn, char* buff, size_t buff_size)
```
dirkschulze: I wrote this some time ago. Looking at it again we should probably use something like [cairo-surface-get-mime-data](https://www.cairographics.org/manual/cairo-cairo-surface-t.html#cairo-surface-get-mime-data).

```c
const unsigned char* data;
unsigned long length;

cairo_surface_get_mime_data (surface, mime_type, &data, &length);
```

dirkschulze: Using size_t instead of unsigned long would be fine with me if that gets accepted by C compilers.

svgnative/example/testC/TestMain.c
```diff
@@ -0,0 +1,102 @@
+/*************************************************************************
+ * ADOBE CONFIDENTIAL
+ * ___________________
+ *
+ * Copyright 2019 Adobe
```
dirkschulze: You probably should add your own (c)

svgnative/example/testC/TestMain.c
```diff
@@ -0,0 +1,102 @@
...
+
+
+    if (argc != 3)
+    {
+        printf("Incorrect number of arguments.");
+        return 0;
+    }
+    buff_input = read_svg_input(argv[1]);
+
+    svg_native_color_map_t* colorMap = svg_native_color_map_create();
```
dirkschulze: Doesn't proper C require instantiation of variables at the beginning of the function even though initialization might happen later?

svgnative/src/SVGNativeCWrapper.cpp
mpsuzuki: I understand you suggest making test program emits rasterized image, instead of text info, I agree, and will do it.

mpsuzuki: At 5288828 , still Text port is selectable, how do you think? Should I remove it completely?

dirkschulze:
@mpsuzuki I think there is a misunderstanding. The Cairo function was an example how to pass the arguments. I didn’t mean to replace the text part with raster content. IMO this should stay the same to keep it simple as an example.

mpsuzuki:
Oh, I'm sorry! I would retry.

mpsuzuki:
I did like this:
```
void svg_native_get_output(svg_native_t* sn, const char** data, size_t* length)
{
    if (!length || !data)
    {
        return;
    }


    *data = NULL;
    *length = 0;


    auto _sn = dynamic_cast<svg_native_t_*>(sn);
    if (!_sn || _sn->mRendererType != SVG_RENDERER_STRING)
        return;


    auto renderer = std::dynamic_pointer_cast<SVGNative::StringSVGRenderer>(_sn->mRenderer);
    if (!renderer)
        return;


    const auto& string = renderer->String();


    *data = string.c_str();
    *length = string.size();


    return;
}
```

dirkschulze:
@mpsuzuki The Argument list looks good now. However, the string just exists as long as the renderer exists and gets cleared together with the renderer. You still want to allocate a limited amount of memory (still 2048 max?) and memcopy the data into the new buffer.

lemzwerg:
Without testing, this looks very nice, thanks!

svgnative/src/SVGNativeCWrapper.cpp
```diff
@@ -239,20 +239,30 @@ void svg_native_render_size(svg_native_t* sn, float width, float height)
...
         return;
+
     const auto& string = renderer->String();
-    auto outputLength = std::min(length, static_cast<int>(string.size()));
-    length = outputLength;
-    for (int i = 0; i < outputLength; i++)
-        output[i] = string[i]; 
+
+    *data = string.c_str();
+    *length = string.size();
```
dirkschulze: As written in another comment, you need to copy the string. You can not just pass the pointer to the string since the string gets cleaned up when the renderer gets cleaned up. Maybe we should still limit the max string size to a certain amount. Could be more than the 2048 it was before though.

mpsuzuki: Umm, at present, my draft is something like this:
```
#ifdef USE_TEXT
void svg_native_get_output(svg_native_t* sn, char** buff, size_t* length)
{
    if (!length)
        return;

    auto _sn = dynamic_cast<svg_native_t_*>(sn);
    if (!_sn || _sn->mRendererType != SVG_RENDERER_STRING)
        return;

    auto renderer = std::dynamic_pointer_cast<SVGNative::StringSVGRenderer>(_sn->mRenderer);
    if (!renderer)
        return;

    const auto& string = renderer->String();

    /*
     * a) buff == NULL: return the length only with no data.
     * b) *buff != NULL: copy the data to preallocated buffer. *length is updated.
     * c) *buff == NULL: allocate new buffer to copy the data.
     */

    /* a) no copy, return the length. */
    if (!buff)
    {
        *length = string.size() + 1;
        return;
    }

    /* b) copy to preallocated buffer */
    if (*buff)
    {
        if (!(*length))
            return;

        memset(buff, 0, *length);
        if (string.size() < *length)
            *length = string.size();
        memcpy(buff, string.c_str(), *length);
        return;
    }

    /* c) copy to new buffer */
    *length = string.size();
    *buff = (char*)malloc(*length);
    memcpy(*buff, string.c_str(), *length);

    return;
}
#endif
```
The reason of complexity is "how to retrieve whole data, and how to deal the buffer allocated by the client (of this library)". I understand this API is not for the real usage but the sample, so I can reduce some features. How do you think of?

dirkschulze: I think we should:
* set `*buff` to `NULL` and `*length` to `0`,
* then we use the `auto size = string.length() + 1;` (should we still limit the max string length? Maybe not.)
* then allocate the buffer `*buf = (const char*) malloc(size * sizeof(const char*));`
* `memcpy(*buff, string.c_str(), size);` (and set the null termination at the end manually or use memset first as you did before.)
* Last but not least set `*length` to `size` and return.

No need to check if the caller created some buffer before. The caller is not supposed to handle this.

dirkschulze: Note, if we do not limit the string length manually, there is no need to use memset or manually set the last char to null terminal. Probably not worth it to keep the limitation for the purpose of demonstrating the interface.

mpsuzuki: Oh, sorry my commit is too late and made a conflict with the symbol export patch. I would do rebase.

svgnative/src/SVGNativeCWrapper.cpp
```diff
@@ -239,20 +239,32 @@ void svg_native_render_size(svg_native_t* sn, float width, float height)
...
-    auto outputLength = std::min(length, static_cast<int>(string.size()));
-    length = outputLength;
-    for (int i = 0; i < outputLength; i++)
-        output[i] = string[i]; 
+    auto size = string.length();
+    *buff = (char*)malloc(size * sizeof(char));
+    if (!(*buff))
+        return;
+
+    memcpy(*buff, string.c_str(), size);
```
dirkschulze: std::string is null terminated by default. You could make  the buffet 1 char bigger and copy one more char. Just make sure you return the correct size to length (actual value of `.size()`)

PR #89, opened 2019-12-29
Do not use dllimport/dllexport for static lib

dirkschulze:
Fix another build issue with clang on windows

### DIFF
--- a/svgnative/include/Config.h
+++ b/svgnative/include/Config.h
@@ -51,9 +51,13 @@ governing permissions and limitations under the License.
 #endif
 
 #ifdef BUILDING_DLL
-	#define SVG_IMP_EXP DllExport
+	#ifdef SVG_IMPORT
+		#define SVG_IMP_EXP DllImport
+	#else
+		#define SVG_IMP_EXP DllExport
+	#endif
 #else
-	#define SVG_IMP_EXP DllImport
+	#define SVG_IMP_EXP
 #endif
 
 #endif // SVGViewer_Config_h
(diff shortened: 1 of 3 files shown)


mpsuzuki:
BTW, the VS project file generated by CMake do not tune the compiler switches for {/MT,/MTd,/MD,/MDd} (= static, static-debug, dynamic, dynamic-debug). In my environment, this switch is always set to MD or MDd. Thus, even if the resulted library is xxx.lib (looking like as a static library), the object files in the library is compiled for shared library.
It is prohibited to link the object files compiled by /MD and those by /MT (for Windows environment - also, the mixture of MDd and MD etc is prohibited). By this limitation, linking with other static library (e.g. libskia.lib) would be prevented, if we use VS project file as it is. If we tweak the compiler switch manually, this problem could be fixed.

Yet I've not tested by myself, CMake after 3.15 has a [MSVC_RUNTIME_LIBRARY](https://cmake.org/cmake/help/latest/prop_tgt/MSVC_RUNTIME_LIBRARY.html#prop_tgt:MSVC_RUNTIME_LIBRARY) to manipulate this.

My idea is:
* if cmake is executed for Windows, and it is newer than 3.15, static/shared option and debug-or-not option sets MSVC_RUNTIME_LIBRARY appropriately.
* if cmake is executed for Windows, but it is older than 3.14, a warn saying "you should tweak project file by yourself" would be given, if the target is static.
How do you think about?

dirkschulze:
@mpsuzuki sounds good to me!

PR #91, opened 2019-12-31
include Win32 headers to compile SVGNativeCWrapper.cpp with GDI+.

mpsuzuki:
It seems that [SVGNativeCWrapper compile error around GDI+ header](https://circleci.com/gh/adobe/svg-native-viewer/655?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) might be caused by the lack of a few header files. I apologize that I could not find this error before.


dirkschulze:
@mpsuzuki no need to apologize! That is why I am adding a Windows build system so that we get notified if we didn't add something.

I noticed that testC still does not build. Actually, I think we don't need the includes you added after all. The GDI include is in the GDIPlusRenderer header already. The issue might be that testC doesn't get build as win32 application. See CMakeFile in testGDIPlus.

Issue #96, opened 2020-01-04
GDI+ and DLL issue


mpsuzuki:
As CircleCI tested [Job 684](https://circleci.com/gh/adobe/svg-native-viewer/684?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link), GDI+ ports have 2 issues.

1.  If -DSHARED=ON, SVGNativeViewerLib.dll could not be built, because the linker cannot find which DLL should be used to resolve GDI+ symbols. It seems that GDI (not GDI+) could be found automatically by default, but GDI+ symbols could not be. To fix this, include "gdiplus.dll" into the linker flag to make SVGNativeViewerLib.dll, like 9d6e874519b6df08d8602edd83ea637ff4980758.

2. It seems that GDI+ header files make the code generator generate the reference to the symbols which are defined but not used. Several GDIPlusSVGRenderer instance methods (they are published) are  implemented within GDIPlusSVGRenderer.h, but uses non-public constructors or destructors (e.g. GDIPlusSVGPath class), and the code generator confused by GDI+ header generates a code referring these non-public constructors. As a result, these object files would have unresolvable symbols. It could be fixed by moving the implementation from GDIPlusSVGRenderer.h to GDIPlusSVGRenderer.cpp, like a9c2f10176e0f2c21820d0a95eede21c56ec342d. We cannot find such issue (on Win32) in Text or Skia backend, so this issue might be caused by Windows header file.

mpsuzuki:
If non-public API should be moved out to separate header file (and it should not be installed, aslike FreeType's internal APIs are defined in internal headers which are not installed by default), 21a87c7eceba54913b1e24ff7bf4bd8c7092f4e5 might be considerable option.

mpsuzuki:
PR #97 is merged, close this.

PR #97, opened 2020-01-06
fix GDIPlus under -DSHARED=ON config

mpsuzuki:
Fix for the issue #96.
* CMakeLists.txt: add explicit dependency to gdiplus.dll.
* GDIPlusSVGRenderer.{h,cpp}: move several inline functions from header to C++ source.
some Windows headers makes the symbol references from the inline functions in the class definition,
even if such functions are not used at all.


mpsuzuki:
Thank you for review! Now I test CircleCI config including "-DGDIPLUS=ON -DSHARED=ON" case.

PR #98, opened 2020-01-06
Initial version or XML wrapper with libxml2 support.

### LINKED ISSUE #17: Allow clients to add different XML parsers
We should allow clients to provide and use different XML parsers (platform provided XML parsers?). This may require a slim wrapper for the XML/DOM function calls in  SVG Native Viewer.

dirkschulze:
Fixes #17 and adds 2 XML parser as example:
* libxml2
* RapidXML (boost)

### DIFF
--- a/svgnative/CMakeLists.txt
+++ b/svgnative/CMakeLists.txt
@@ -381,7 +381,12 @@ endif()
 if (LIBXML2_FOUND)
     target_link_libraries(SVGNativeViewerLib "${LIBXML2_LIBRARIES}")
     target_include_directories(SVGNativeViewerLib PUBLIC  "${LIBXML2_INCLUDE_DIR}")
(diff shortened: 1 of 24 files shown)


dirkschulze:
This patch does various optimizations and fixes to compensate the bigger lib size for the XML wrapper. The performance was improved too. The wrapper has no negative impact on performance.

dirkschulze:
@qrb7xc It would be great if you could take a look as well. If there is anything I can do to make the wrapper thinner or more efficient, please let me know.

svgnative/CMakeLists.txt
```diff
@@ -113,6 +115,8 @@ else()
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -pedantic -Wextra")
 endif()
 
+#set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g")
```
dirkschulze: will remove

mpsuzuki: good!

PR #99, opened 2020-01-07
Use constants for all strings

dirkschulze:
Collecting all constant strings in one header as constexpr. I am experimenting with MSXML that operates on UTF16. This might make it easier to use MSXML by choosing a different char encoding internally.


svgnative/src/SVGDocumentImpl.cpp
```diff
@@ -166,8 +168,8 @@ void SVGDocumentImpl::ParseChild(const XMLNode* child)
     }
 
     // Look at all elements that are no shapes.
-    std::string elementName = child->GetName();
-    if (elementName == "g")
+    const auto elementName = child->GetName();
```
dirkschulze: Smaller performance improvements by avoiding copies.

svgnative/src/SVGDocumentImpl.cpp
```diff
@@ -574,26 +576,28 @@ GraphicStyleImpl SVGDocumentImpl::ParseGraphic(
     return graphicStyle;
 }
 
+static inline void AddDetectedProperty(const XMLNode* node, PropertySet& propertySet, const char* propertyName)
```
dirkschulze: Turn lambda to inline function.

PR #102, opened 2020-01-10
Enable Skia port on non-Apple platforms.

mpsuzuki:
Currently, building Skia port is restricted to Apple platform. This patch enables Skia on non-Apple platforms. The location and library file format are platform dependent, I introduced 2 option variables to be set by users.
* SKIA_LIBRARY_PATH
full-path specification of the skia library to be linked. if the user want to link skia.dll, SKIA_LIBRARY_PATH should point skia.dll.lib.
* WIN32_OBJECT_TYPE
the option to be passed to MSVC or Clang-CL, to control the object type as sharable or non-sharable ("/MD" or "/MT"). This has no effect on non-Windows platforms. 

The behaviour without setting them is same with non-patched CMakeLists.txt.

Testing of patched source tree is being done at [CircleCI](https://circleci.com/gh/mpsuzuki/test-circleci-2020/29).

If you think such detailed CMake scripting should be out of main CMakeLists.txt and placed in cmake/ folder (as FindCairo.cmake module), please let me know.


mpsuzuki:
Oops, checking of the version of cmake does not work well. I must update.

svgnative/CMakeLists.txt
```diff
@@ -95,7 +128,7 @@ endif()
 if(EXISTS "${CMAKE_TOOLCHAIN_FILE}")
     # do not apply any global settings if the toolchain
     # is being configured externally
-elseif(MSVC)
+elseif(WIN32)
```
dirkschulze: I am not sure if that is correct. The following settings are specific to VS. Have you tried building with VS+clang or migw?

svgnative/CMakeLists.txt
```diff
@@ -52,19 +56,48 @@ CMAKE_DEPENDENT_OPTION(USE_PLATFORM_XML "Platform XML parser" TRUE "PLATFORM_XML
...
+# setting for Skia
+################################
+if(SKIA)
+    if (NOT DEFINED SKIA_LIBRARY_PATH)
+        if (APPLE)
+            set(SKIA_LIBRARY_PATH "${CMAKE_CURRENT_SOURCE_DIR}/../third_party/skia/lib/osx/libskia.a")
+        elseif (WIN32)
+            set(SKIA_LIBRARY_PATH "${CMAKE_CURRENT_SOURCE_DIR}/../third_party/skia/lib/skia.lib")
+        else()
+            message(FATAL_ERROR "Fatal error: No usable Skia library is found, please specify -DSKIA_LIBRARY_PATH=XXX.")
```
dirkschulze: was found

dirkschulze:
> If you think such detailed CMake scripting should be out of main CMakeLists.txt and placed in cmake/ folder (as FindCairo.cmake module), please let me know.

Maybe we should, yes.

svgnative/CMakeLists.txt
mpsuzuki: OK, I would revert them to MSVC (I was thinking about Clang, but overlooked about MinGW). I already confirmed clang-cl.exe toolset (provided as a component of MSVS) can build the set, by changing the toolset from MSVS GUI. It seems that changing the toolset on Windows commandline is quite complicated (please find [How to use clang-cl and lld-link with Visual Studio generator](https://gitlab.kitware.com/cmake/cmake/issues/19174)). I would postpone it as future issue.

svgnative/CMakeLists.txt
mpsuzuki: Oh, thanks!

mpsuzuki:
I moved the part for Skia library pathname & compiler flag to cmake/FindSkia.cmake. But I could not move the part to changing the compiler flag (for MSVC) to separated file (my attempt is like this 9ed60cc8a27786573ef3c4f945a49d0b93843f88). It seems that MSVC runtime library hook does not work well within the user-defined functions.

mpsuzuki:
For Win32 testing of Skia, I use following items.
- My Skia m70 headers & binaries (shared/static) at [github.com/mpsuzuki/skia-binaries](https://github.com/mpsuzuki/skia-binaries/releases/download/m70-win64-20191207/skia-m70-clang-on-win64_20191207-1815.tar.xz)
- Boost headers required for svg-native-viewer at [my office](https://home.hiroshima-u.ac.jp/mpsuzuki/boost-1.72.0_required-by-svg-native-viewer.7z). By using this, the installation process of Boost would be minimized, the consumed time is effectively reduced from 6-8min to 10-30sec. The developers using "free plan" of CircleCI, it might useful.

For Linux, I was using headers & binaries at:
- My Skia m70 headers & binaries built on Ubuntu at [my office](https://home.hiroshima-u.ac.jp/mpsuzuki/skia-m70_ubuntu1804_20190517.cpio.xz)
But, the different dependencies between Ubuntu and Debian is not easy for most developers (to use this binary, many library-development files should be installed). I should compile another set with smaller dependencies.

svgnative/cmake/FindSkia.cmake
```diff
@@ -0,0 +1,88 @@
+# - Try to find the Skia library to be linked with,
+# and prepare a few variables for linker flags.
+#
+# ----------------------------------------------------------------------------
+#
+# Copyright (C) 2020 suzuki toshiya <mpsuzuki@hiroshima-u.ac.jp>
+#
+# Redistribution and use in source and binary forms, with or without
```
dirkschulze: Do you mind putting this under Apache License?

mpsuzuki: OK!

mpsuzuki:
Umm, installation of genuine & fullsize Boost exceeded 10 min, and CI process is aborted. I would retry with rebased version, after PR#105 is merged.

mpsuzuki:
By merged PR #105, the CI time for Windows is significantly reduced! Please let me merge.

PR #107, opened 2020-01-23
TestMain.c: change the license to APL 2.0

mpsuzuki:
Maybe it was my mistake that "Adobe Confidential" license is used in TestMain.c.
Using Apache License 2.0 would be better and consistent.


dirkschulze:
@mpsuzuki Thanks! Sorry for the long delay. Was busy lately :(.

Issue #111, opened 2020-05-13
 This image cannot be displayed.


jonas-imagine:
``` 
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="150" height="250" viewBox="0 0 150 250">
    <linearGradient id="g1" gradientUnits="userSpaceOnUse">
        <stop stop-color="#00FFFF" offset="0"/>
        <stop stop-color="#00FF00" offset="1"/>
    </linearGradient>
    <g fill="none">
    	<rect width="150" height="10" fill="url(#g1)" y="180"/>
    </g>
</svg>
```

Works when removing fill="none".

mpsuzuki:
@jonas-imagine, I apologize my delayed response. I confirmed your error by testText/testSVGNative testing program (reading SVG and emitting a text to show the internal recognization of the SVG input). The difference of the original SVG (nothing shown) and the modified SVG (remove 'fill="none"', showing a rectangle) is something like:
```
--- gh-issue111bad.txt  2020-05-18 11:14:34.287798322 +0900
+++ gh-issue111good.txt 2020-05-18 11:14:39.263946146 +0900
@@ -2,7 +2,7 @@
     [group
         [group
             [path Rect(0,180,150,10)
-                fill: {hasFill: false winding: nonzero paint: {
+                fill: {hasFill: true winding: nonzero paint: {
                     linearGradient: x1: 0 y1: 0 x2: 150 y2: 0 method: pad stops: {
                         offset: 0 rgba(0,1,1,1)
                         offset: 1 rgba(0,1,0,1)
```
As you can see, the fill="url(#1)" is correctly parsed, but the "hasFill" boolean (of the internal data in SVG Native Viewer) is incorrectly(?) inherited from the parent g element.

I think this difference is caused by SVGStringParser::ParsePaint(), which sets the default return value to SVGDocumentImpl::Result::kInvalid, but never update it to kSuccess, even if the parsing process works correctly.

Maybe I can draft to fix this issue soon, but I'm unsure whether it is acceptable with the original implementation idea.

jonas-imagine:
Has solved my problem.👍

mpsuzuki:
Thank you for testing! I'm waiting for the review & merge by @dirkschulze... If you have any urgency for this issue, please let me know.

mpsuzuki:
[...]

PR #112, opened 2020-05-18
Fix for Issue #111: a valid fill attribute cannot override its parental "fill='none'" attribute.

mpsuzuki:
SVGStringParser::ParsePaint() sets the initial value of 'result' to kInvalid. In some cases, the result is updated to kDisabled, but it is never updated to kSuccess, except of the case that ParseColor() updates it. Therefore, once the parental element sets "fill='none'", often the child elements cannot override it, like Issue #111.

This commit changes the initial value from kInvalid to kSuccess, and all 'return result' before changing 'result' are replaced by immediate 'return kInvalid'.


dirkschulze:
Thanks a lot @mpsuzuki !

mpsuzuki:
@dirkschulze , sorry for delayed action! To work with other issues #114, #115, I've committed this change set.

PR #113, opened 2020-07-23
Update to latest skia and some improvements

sunflowerdeath:
Hello, I wanted to use your library but I have latest stable version of skia (m85), so I updated the code. It required me to enable  c++14 and to fix some minor API changes. Also, I added cmake flag `SKIA_SOURCE_DIR` that is similar to  `SKIA_LIBRARY_PATH`, that allows to specify location of Skia's sources when it is not in the `third_party` directory.

Also, I moved all public headers (including ports) into `include/svgnative` directory. This is good practice that is used by many modern cpp libraries, for example: https://github.com/fmtlib/fmt/tree/master/include/fmt, https://github.com/glfw/glfw/tree/master/include/GLFW, https://github.com/gabime/spdlog/tree/v1.x/include/spdlog
This helps to understand to which library header belong and helps avoid name conflicts (`#include "SVGDocument.h"` => `#include "svgnative/SVGDocument.h"`).


svgnative/CMakeLists.txt
```diff
@@ -16,10 +16,10 @@ if (CMAKE_VERSION VERSION_GREATER "3.15")
     cmake_policy(SET CMP0091 NEW)
 endif()
 
-set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
+set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 project(SVGNativeViewer)
 enable_language(CXX)
-set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD 14)
```
dirkschulze, line 22: It is a project requirement, due to other projects, to support C++ 11. Sadly, we can not change it to 14 at this point.

svgnative/CMakeLists.txt
```diff
@@ -234,17 +234,17 @@ file(GLOB gl_headers ${PROJECT_SOURCE_DIR}/include/*.h)
 set(text_port)
 if(USE_TEXT)
 file(GLOB text_port
-    ports/string/StringSVGRenderer.h
-    ports/string/StringSVGRenderer.cpp
+    src/ports/string/StringSVGRenderer.h
+    src/ports/string/StringSVGRenderer.cpp
```
dirkschulze, line 238: Leaving the ports separate of the actual source was a requirement by other projects. I agree that it is odd, but I need to check if these requirements are still relevant. Could you leave the ports/ folder in root for now?

svgnative/example/testC/TestMain.c
```diff
@@ -17,7 +17,7 @@ governing permissions and limitations under the License.
 #include <stdio.h>
 #include <stddef.h>
 
-#include "SVGNativeCWrapper.h"
+#include "svgnative/SVGNativeCWrapper.h"
```
dirkschulze, line 20: I agree that it is a good idea to move the includes.

dirkschulze (changes requested):
Thanks a lot for your contributions @sunflowerdeath and sorry for the delay in review. I like the changes. There are some requirements force by other projects that make it difficult to do some of the file moves. See inline comments.

dirkschulze:
@sunflowerdeath I see that Skia indeed requires a C++14 compatible compiler. IMO we should only require C++14 when the Skia port is enabled.

You patch looks great otherwise. Let me know if you get the time to do the necessary change. Otherwise I can do it on top of your changes as well.

sunflowerdeath:
> we should only require C++14 when the Skia port is enabled.
This is very good idea.

> Let me know if you get the time to do the necessary change. Otherwise I can do it on top of your changes as well.
Currently I might not have time for it, so if you could do it, it woud be great.

dirkschulze:
Covered by #118 now. @sunflowerdeath, could you take a look at the changes on that PR please?

Issue #117, opened 2020-08-26
SVGStringParser.cpp -- a sample SVG path that SVG Native interprets differently from other SVG viewers


dejanyy:
Attached is a sample and stripped down SVG file that SVG Native seems to interpret (and render) differently from all other SVG viewers I tired (e.g. Firefox, Chrome, Safari, Apple's Preview). All these SVG renders show the attached SVG file as a glyph "K". But with SVG Native, the top diagonal line in K is distorted.

I've analyzed the code in SVGStringParser.cpp (around lines 385 - 456) and, it appears to me, that something is not quite right with path processing logic in the ParsePathString function. Maybe I don't fully understand this code, however, it seems a bit strange to me how the startSet boolean variable is used.

I was able to make some very small changes to this code (i.e. modified just a few lines) to basically remove dependence on the startSet variable. With these changes made, the attached SVG file renders exactly the same as with other SVG viewers I tested.

If you are able to confirm that SVG Native is not interpreting this particular SVG file correctly, would you be interested in receiving and reviewing my changes?

[strange_looking_k.svg.txt](https://github.com/adobe/svg-native-viewer/files/5132331/strange_looking_k.svg.txt)

dejanyy:
One more comment. In the above SVG file, the glyph id is associated with the root <svg> element:

<svg id="glyph53" ...
<g><path ... ></g>
</svg>

I believe this is contrary to the OpenType SVG spec (judging by the "Glyph Identifiers" section at https://docs.microsoft.com/en-us/typography/opentype/spec/svg), but who knows...

It looks like SVG Native can't find glyph ids that are attached to the root svg element.

However, the above SVG file is extracted from a font called BungeeColor-Regular_svg.ttf, so it appears there are fonts out there that like to attach glyph ids to the root svg element.

dirkschulze:
@dejanyy Thanks a lot for reporting this inconsistency. Regardless of the spec, in this project we generally try to align with browser implementations. If all browsers are rendering it consistently, then the current implementation of svg-native-viewer should change.

I'd love to review a patch. Maybe you could elaborate what is incorrect on the use of `startSet`?

Please open another issue for the ID of the root element. I am currently not sure if that ID should match or not but would check.

dejanyy:
[...]

PR #124, opened 2020-11-12
A moveTo should reset the last current position but didn't. Issue #117

dirkschulze:


svgnative/src/SVGStringParser.cpp
```diff
@@ -403,8 +412,14 @@ void ParsePathString(const std::string& pathString, Path& p)
             return;
         if (!isDigit(*pos) && *pos != ',' && *pos != '-' && *pos != '.')
             prev = *pos++;
-        else if (prev == 'z' || prev == 'Z')
+        else if (IsCloseCmd(prev))
             break;
+        // Do not close a path if we do not have one yet.
+        if (IsCloseCmd(lastCommand) && IsCloseCmd(prev))
+            continue;
```
qrb7xc, line 419: So prev is ignored in this case? Shouldn't we warn about this?

qrb7xc (approved):

svgnative/src/SVGStringParser.cpp
dirkschulze: In SVG it is valid to have multiple close paths commands w/o creating a new subpaths. This change makes sure we don't call closePath on the renderer multiple times.

PR #133, opened 2021-03-16
D2D port fixes

qrb7xc:
Fix the over-release issue with a ID2D1PathGeometry object and update the test app from a D2D test sample source.

The over-release issue can be fixed by avoiding a copy of an D2DSVGPath but to make D2D objects generally copyable and handle lifetime issues all D2D objects are now wrapped by a smart pointer (CComPtr). The only downside here is that CComPtr is part of Windows' Active Template Library (ATL) which might not be available everywhere.

### DIFF
--- a/svgnative/example/testD2D/TestD2D.cpp
+++ b/svgnative/example/testD2D/TestD2D.cpp
@@ -22,180 +22,180 @@ governing permissions and limitations under the License.
 
 template <class T> void SafeRelease(T** ppT)
 {
-	if (*ppT)
-	{
-		(*ppT)->Release();
-		*ppT = nullptr;
-	}
+    if (*ppT)
+    {
+        (*ppT)->Release();
+        *ppT = nullptr;
+    }
 }
 
 namespace
 {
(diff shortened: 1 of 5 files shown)


svgnative/example/testD2D/TestD2D.cpp
```diff
@@ -12,116 +12,190 @@ governing permissions and limitations under the License.
...
 
+#include "basewin.h"
 #include "svgnative/SVGDocument.h"
 #include "svgnative/ports/d2d/D2DSVGRenderer.h"
 
-/* We need to include the Gdiplus lib */
-#pragma comment (lib, "D2d1.lib")
+template <class T> void SafeRelease(T** ppT)
+{
+	if (*ppT)
```
dirkschulze, line 25: The rest of the code uses spaces as indentation (actually, I thought we have code style checks in place but maybe the bots are only checking PRs on main.)

svgnative/example/testD2D/TestD2D.cpp
```diff
@@ -12,116 +12,190 @@ governing permissions and limitations under the License.
...
-
-        TranslateMessage(&msg);
-        DispatchMessage(&msg);
-    }
-
-    /* Return exit code of WM_QUIT */
-    return (int)msg.wParam;
+	switch (uMsg)
+	{
+	case WM_CREATE:
```
dirkschulze, line 174: more indentation issues.

dirkschulze (commented):
LGTM. Just a few indentation style changes needed.

dirkschulze:
Code looks good to me an I am going to merge it. Could you work on the CircleCI integration as well or shall I @qrb7xc ?

PR #136, opened 2021-04-24
Update to latest Skia APIs

moazin:
It seems that the current Skia port doesn't work anymore due to changes to Skia's API. These changes worked for me. Please review.


svgnative/CMakeLists.txt
```diff
@@ -19,7 +19,7 @@ endif()
 set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
 project(SVGNativeViewer)
 enable_language(CXX)
-set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD 14)
```
dirkschulze, line 22: This should only be done when Skia is enabled. Also, I wonder if we didn't make the change earlier? I thought that we did.

svgnative/ports/skia/SkiaSVGRenderer.cpp
```diff
@@ -286,8 +288,9 @@ void SkiaSVGRenderer::DrawImage(
     SVG_ASSERT(mCanvas);
     Save(graphicStyle);
     mCanvas->clipRect({clipArea.x, clipArea.y, clipArea.x + clipArea.width, clipArea.y + clipArea.height}, SkClipOp::kIntersect);
+    SkSamplingOptions sso;
```
dirkschulze, line 291: Could we name is samplingOptions for better understanding?

dirkschulze (commented):
Please take a look at the CMake settings.

svgnative/CMakeLists.txt
moazin: Yes, I also just checked. You guys did make some of the changes here. My branch was a few commits behind main. :laughing:  Fixing it.

moazin: Do you think this can be done in the `setting for Skia` block? Or should we NOT set the CXX standard if `CMAKE_TOOLCHAIN_FILE` exists?

dirkschulze: In the Skia settings should be fine. In all other cases we should build with 11.

dirkschulze (approved):
Changes look good to me! Thanks a lot.

moazin:
@dirkschulze You can merge this one too. Thanks for reviewing. 👍

Issue #138, opened 2021-04-26
Path clipping and opacity support is missing in GDIPlus port


moazin:
The GDIPlus port doesn't support SVG Path clipping and opacity. These are marked as TODOs. Thought I'd open issues for them here for record-keeping and as a reminder that this needs to be done.

PR #143, opened 2021-04-28
Fixes issues with clipping, opacity and transforms with GDI+ port

### LINKED ISSUE #140: Gdiplus port doesn't support transforms on elements
Due to a minor coding error, it seems that the Gdiplus port doesn't support transforms on elements.

It seems that in the `GDIPlusSVGRenderer::DrawPath` method, it's missing calls to `::Save` and `::Restore` and is instead directly calling the underlying Save and Restore functions of Gdiplus.

I'll submit a PR soon.

moazin:
This should fix #140 and #138.

### DIFF
--- a/svgnative/src/ports/gdiplus/GDIPlusSVGRenderer.cpp
+++ b/svgnative/src/ports/gdiplus/GDIPlusSVGRenderer.cpp
@@ -13,8 +13,8 @@ governing permissions and limitations under the License.
 #include <Windows.h>
 #include <gdiplus.h>
 
-#include "Config.h"
-#include "GDIPlusSVGRenderer.h"
+#include "svgnative/Config.h"
+#include "svgnative/ports/gdiplus/GDIPlusSVGRenderer.h"
 #include "base64.h"
 
 namespace SVGNative
--- a/svgnative/src/ports/gdiplus/GDIPlusSVGRenderer.cpp
+++ b/svgnative/src/ports/gdiplus/GDIPlusSVGRenderer.cpp
@@ -230,7 +230,7 @@ void GDIPlusSVGRenderer::Save(const GraphicStyle& graphicStyle)
             clip_path->Transform(matrix);
         }
         clip_path->SetFillMode(graphicStyle.clippingPath->clipRule == WindingRule::kNonZero ? Gdiplus::FillMode::FillModeWinding : Gdiplus::FillMode::FillModeAlternate);
-        mContext->SetClip(clip_path.get());
+        mContext->SetClip(clip_path.get(), Gdiplus::CombineModeIntersect);
     }
 }
 
--- a/svgnative/src/ports/gdiplus/GDIPlusSVGRenderer.cpp
+++ b/svgnative/src/ports/gdiplus/GDIPlusSVGRenderer.cpp
@@ -13,8 +13,8 @@ governing permissions and limitations under the License.
 #include <Windows.h>
 #include <gdiplus.h>
 
-#include "svgnative/Config.h"
-#include "svgnative/ports/gdiplus/GDIPlusSVGRenderer.h"
+#include "Config.h"
+#include "GDIPlusSVGRenderer.h"
 #include "base64.h"
 
 namespace SVGNative
@@ -223,14 +223,14 @@ void GDIPlusSVGRenderer::Save(const GraphicStyle& graphicStyle)
 
     if (graphicStyle.clippingPath)
     {
-        // TODO
-        SVG_ASSERT_MSG(false, "Need to implement svg clipping path!");
-    }
-
-    if (graphicStyle.opacity < 1.0f)
-    {
-        // TODO
-        SVG_ASSERT_MSG(false, "Need to implement svg opacity!");
+        std::unique_ptr<Gdiplus::GraphicsPath> clip_path(dynamic_cast<const GDIPlusSVGPath*>(graphicStyle.clippingPath->path.get())->GetGraphicsPath().Clone());
+        if (graphicStyle.clippingPath->transform)
+        {
+            const Gdiplus::Matrix* matrix = &dynamic_cast<GDIPlusSVGTransform*>(graphicStyle.clippingPath->transform.get())->GetMatrix();
+            clip_path->Transform(matrix);
+        }
+        clip_path->SetFillMode(graphicStyle.clippingPath->clipRule == WindingRule::kNonZero ? Gdiplus::FillMode::FillModeWinding : Gdiplus::FillMode::FillModeAlternate);
+        mContext->SetClip(clip_path.get());
     }
 }
 
@@ -328,7 +328,7 @@ std::unique_ptr<Gdiplus::Brush> GDIPlusSVGRenderer::CreateGradientBrush(const Gr
 void GDIPlusSVGRenderer::DrawPath(const Path& renderPath, const GraphicStyle& graphicStyle, const FillStyle& fillStyle, const StrokeStyle& strokeStyle)
 {
     SVG_ASSERT(mContext);
-    Gdiplus::GraphicsState savedState = mContext->Save();
+    Save(graphicStyle);
 
     std::unique_ptr<Gdiplus::GraphicsPath> path(dynamic_cast<const GDIPlusSVGPath&>(renderPath).GetGraphicsPath().Clone());
     if (fillStyle.fillRule == WindingRule::kEvenOdd)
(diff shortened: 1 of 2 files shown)


svgnative/src/ports/gdiplus/GDIPlusSVGRenderer.cpp
```diff
@@ -223,14 +223,14 @@ void GDIPlusSVGRenderer::Save(const GraphicStyle& graphicStyle)
...
-        // TODO
-        SVG_ASSERT_MSG(false, "Need to implement svg opacity!");
+        std::unique_ptr<Gdiplus::GraphicsPath> clip_path(dynamic_cast<const GDIPlusSVGPath*>(graphicStyle.clippingPath->path.get())->GetGraphicsPath().Clone());
+        if (graphicStyle.clippingPath->transform)
+        {
+            const Gdiplus::Matrix* matrix = &dynamic_cast<GDIPlusSVGTransform*>(graphicStyle.clippingPath->transform.get())->GetMatrix();
+            clip_path->Transform(matrix);
+        }
+        clip_path->SetFillMode(graphicStyle.clippingPath->clipRule == WindingRule::kNonZero ? Gdiplus::FillMode::FillModeWinding : Gdiplus::FillMode::FillModeAlternate);
+        mContext->SetClip(clip_path.get());
```
dirkschulze, line 233: It is quite some time ago that I looked into GDI+. IIRC, clipping worked differently with GDI+ than with other rendering engines. Have you tried with SVG files where a (transformed?) group with nested graphical elements/shapes gets clipped?

```xml
<clipPath id="clip">
    <ellipse rx="25" ry="50" cx="100" cy="100"/> 
</clipPath>
<g transform="rotate(45)" clip="url(#clip)">
    <g clip="url(#clip)">
        <rect fill="green" width="200" height="200"/>
    </g>
</g>
```

svgnative/src/ports/gdiplus/GDIPlusSVGRenderer.cpp
```diff
@@ -344,7 +344,7 @@ void GDIPlusSVGRenderer::DrawPath(const Path& renderPath, const GraphicStyle& gr
         if (fillStyle.paint.type() == typeid(Color))
         {
             auto color = boost::get<Color>(fillStyle.paint);
-            color[3] *= fillStyle.fillOpacity;
+            color[3] *= fillStyle.fillOpacity * graphicStyle.opacity;
```
dirkschulze, line 347: Sure that we don't need to multiply all opacity values of previous states? Was thinking of something like:

```xml
<g opacity="0.5">
    <g opacity="0.5">
        <rect fill-opacity="0.5" fill="green"/>
    </g>
</g>
```

dirkschulze (commented):
The changes look good to me in general. Just wonder if they work for nested graphics. See comments and examples in the review.

svgnative/src/ports/gdiplus/GDIPlusSVGRenderer.cpp
```diff
@@ -223,14 +223,14 @@ void GDIPlusSVGRenderer::Save(const GraphicStyle& graphicStyle)
...
-        // TODO
-        SVG_ASSERT_MSG(false, "Need to implement svg opacity!");
+        std::unique_ptr<Gdiplus::GraphicsPath> clip_path(dynamic_cast<const GDIPlusSVGPath*>(graphicStyle.clippingPath->path.get())->GetGraphicsPath().Clone());
+        if (graphicStyle.clippingPath->transform)
+        {
+            const Gdiplus::Matrix* matrix = &dynamic_cast<GDIPlusSVGTransform*>(graphicStyle.clippingPath->transform.get())->GetMatrix();
+            clip_path->Transform(matrix);
+        }
+        clip_path->SetFillMode(graphicStyle.clippingPath->clipRule == WindingRule::kNonZero ? Gdiplus::FillMode::FillModeWinding : Gdiplus::FillMode::FillModeAlternate);
+        mContext->SetClip(clip_path.get(), Gdiplus::CombineModeIntersect);
```
moazin, line 233: @dirkschulze Thanks for pointing out about clipping working differently with Gdiplus. That's true. In fact, it allows you to specify how you want to form the new clipping region. Other rendering engines such as Cairo form the new clipping region by intersection the passed path with the old one and by passing this flag, GDI+ should do the same. The SVG document you provided gives the following rendering:
image
Chrome seems to render it the same way. Is this the expected output?

svgnative/src/ports/gdiplus/GDIPlusSVGRenderer.cpp
moazin: Oh you're right. Thanks for pointing this out. Bascially, Skia and CG have mechanisms for transparency layers which automatically do this. But GDI+ and Cairo don't. So I'm thinking we can stack these values in a `std::stack`?

svgnative/src/ports/gdiplus/GDIPlusSVGRenderer.cpp
```diff
@@ -223,21 +223,27 @@ void GDIPlusSVGRenderer::Save(const GraphicStyle& graphicStyle)
...
 
-    if (graphicStyle.opacity < 1.0f)
-    {
-        // TODO
-        SVG_ASSERT_MSG(false, "Need to implement svg opacity!");
-    }
+    float opacity = 1.0;
+    if (mOpacityStack.size() > 0)
+        opacity = mOpacityStack.top();
+    mOpacityStack.push(opacity * graphicStyle.opacity);
```
moazin, line 239: @dirkschulze If this seems the right way to do it, I can replicate the same in Cairo. Rendering results seem accurate to me.

svgnative/src/ports/gdiplus/GDIPlusSVGRenderer.cpp
moazin: @dirkschulze I just realized that the transformation in `::Save` in the Gdiplus port is messed up too. It needs to be concatenated and the way it's happening right now, the matrix is being set directly.

svgnative/src/ports/gdiplus/GDIPlusSVGRenderer.cpp
dirkschulze: @moazin Cairo does have a layer like concept that can be used for things like raster masks. For opacity and and clipping the internal context stack is good enough (save/restore logic).

For GDI+ yes, we would need to multiple the opacity values of previous save() operations. However, not sure what can be done for clipping. I actually don't know enough about GDI+ to give advices for clipping.

moazin: @dirkschulze Have a look at the [documentation](https://docs.microsoft.com/en-us/windows/win32/api/gdiplusgraphics/nf-gdiplusgraphics-graphics-setclip(constgraphicspath_combinemode)) for `::SetClip`. It has a `CombineMode` parameter and setting it to intersect should solve the problem as far as I can see.

svgnative/src/ports/gdiplus/GDIPlusSVGRenderer.cpp
dirkschulze: @moazin Cairo graphics uses a state machine and has a save/restore logic. For simple opacity/clipping there is no additional logic needed. No need to multiply with ancestor opacities.

svgnative/src/ports/gdiplus/GDIPlusSVGRenderer.cpp
```diff
@@ -217,27 +217,33 @@ void GDIPlusSVGRenderer::Save(const GraphicStyle& graphicStyle)
...
-            mContext->SetTransform(matrix);
+            mContext->MultiplyTransform(matrix, Gdiplus::MatrixOrderPrepend);
         }
     }
 
     if (graphicStyle.clippingPath)
     {
-        // TODO
-        SVG_ASSERT_MSG(false, "Need to implement svg clipping path!");
+        std::unique_ptr<Gdiplus::GraphicsPath> clip_path(dynamic_cast<const GDIPlusSVGPath*>(graphicStyle.clippingPath->path.get())->GetGraphicsPath().Clone());
```
dirkschulze, line 226: Please do not mix camelCase and snake-case here.

dirkschulze (approved):

svgnative/src/ports/gdiplus/GDIPlusSVGRenderer.cpp
dirkschulze: I still wonder if it works with putting transforms on groups with clipping set before and after the transform. The PR still is a big improvement to the current behavior.

moazin: @dirkschulze What do you mean by clipping set before and after the transform? Can you provide an example that I can test?

PR #145, opened 2021-04-30
Fix the rotation problem in CG port

### LINKED ISSUE #144: CoreGraphics port messes up rotations
The CoreGraphics port's rotation transform is messed up because the underlying port call uses radians while the function arguments are in degrees and the conversion needed is missing.

moazin:
Fixes #144. The bug was simple, the angle being passed in to `CGAffineTransformRotate` is in degrees when it should be in radians.


dirkschulze (approved):
Change looks good to me. Thanks a lot!

moazin:
@dirkschulze You can merge this if the change looks good. Thanks for reviewing. 👍

PR #152, opened 2021-05-24
Fix Quadratic Beziers problem

### LINKED ISSUE #151: Quadratic Beziers are not handled properly
Quadratic Beziers are not rendered properly. See the following document and the following renderings:
```
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 1000 1000" width="1000" height="1000">
  <path d="M 100,100 Q 300,100 300,300 L 100,300 Z" stroke="red" stroke-width="10" />
  <path d="M 100,500 Q 200,400 300,500 T 500,500" stroke="red" stroke-width="10" />
  <path d="M 100,700 C 133,600 166,600 200,700 S 266,800 300,700 S 366,600 400,700" stroke="red" stroke-width="10" />
</svg>
```
All three of the following renderings are different. It might look as if Skia agrees with Librsvg, it doesn't if you compare closely.
Rendering by Librsvg:
image
Rendering by SNV (Cairo):
image
Rendering by SNV (Skia):
image

moazin:
This should fix #151. Surprisingly, the GDI+ port already had the necessary conversion code while Cairo didn't. Skia and CG both have API calls for Quadratic Curves and yet they were not being used.

### DIFF
--- a/svgnative/src/ports/cairo/CairoSVGRenderer.cpp
+++ b/svgnative/src/ports/cairo/CairoSVGRenderer.cpp
@@ -122,10 +122,10 @@ void CairoSVGPath::CurveTo(float x1, float y1, float x2, float y2, float x3, flo
 
 void CairoSVGPath::CurveToV(float x2, float y2, float x3, float y3)
 {
-    float cx1 = static_cast<float>(mCurrentX + 2.0 / 3.0 * (x2 - mCurrentX));
-    float cy1 = static_cast<float>(mCurrentY + 2.0 / 3.0 * (y2 - mCurrentY));
-    float cx2 = static_cast<float>(x3 + 2.0 / 3.0 * (x2 - x3));
-    float cy2 = static_cast<float>(y3 + 2.0 / 3.0 * (y2 - y3));
+    float cx1 = mCurrentX + 2.0f / 3.0f * (x2 - mCurrentX);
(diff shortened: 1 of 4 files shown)


svgnative/src/ports/cairo/CairoSVGRenderer.cpp
```diff
@@ -122,7 +122,12 @@ void CairoSVGPath::CurveTo(float x1, float y1, float x2, float y2, float x3, flo
 
 void CairoSVGPath::CurveToV(float x2, float y2, float x3, float y3)
 {
-    cairo_curve_to(mPathCtx, mCurrentX, mCurrentY, x2, y2, x3, y3);
+    float cx1 = static_cast<float>(mCurrentX + 2.0 / 3.0 * (x2 - mCurrentX));
```
dirkschulze, line 125: `2.0f` and `3.0f`. After that, it should not be needed to cast to float.

dirkschulze (approved):

svgnative/src/ports/cairo/CairoSVGRenderer.cpp
moazin: Fixed. Please check, thanks.

dirkschulze (approved):

dirkschulze:
LGTM @moazin . Thanks a lot.

Issue #134, opened 2021-04-05
Arcs with rx and ry being different are not considered invalid


moazin:
Inside the `SVGParserArcToCurve.cpp` on line `109` there is an `SVG_ASSERT` that ensures that `rx` and `ry` are not different. In the event that they are, an assertion is violated, thus the program stops. The comment there says `untested` and I think that means the code hasn't been tested for cases where they are different. The correct behavior is for such arcs to be rendered properly.
```cpp
    SVG_ASSERT(angle == 0); // untested

    if (radiusX == 0 || radiusY == 0)
    {
        // this is actually a line
        path.LineTo(endX, endY);
        return;
    }

    SVG_ASSERT(radiusX == radiusY); // untested

    SVG_ASSERT(radiusX != 0 && radiusY != 0);
```

moazin:
I did some testing and it seems that there are problems with Arcs that have a non-zero angle.
```
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 2000 1000" width="2000" height="1000">
  <path d="M 100,100 a 100 50 -45 0 0 100,0" stroke="red" stroke-width="10" />
  <path d="M 100,200 a 100 50 -45 1 0 100,0" stroke="red" stroke-width="10" />
  <path d="M 250,100 a 100 50 -45 0 1 100,0" stroke="red" stroke-width="10" />
  <path d="M 250,300 a 100 50 -45 1 1 100,0" stroke="red" stroke-width="10" />
  <path d="M 500,100 a 100 50 0 0 0 100,0" stroke="red" stroke-width="10" />
  <path d="M 500,200 a 100 50 0 1 0 100,0" stroke="red" stroke-width="10" />
  <path d="M 650,100 a 100 50 0 0 1 100,0" stroke="red" stroke-width="10" />
  <path d="M 750,300 a 100 50 0 1 1 100,0" stroke="red" stroke-width="10" />
  <path d="M 100,400 a 100 50 90 0 0 100,0" stroke="red" stroke-width="10" />
  <path d="M 100,550 a 100 50 90 1 0 100,0" stroke="red" stroke-width="10" />
  <path d="M 250,450 a 100 50 90 0 1 100,0" stroke="red" stroke-width="10" />
  <path d="M 250,600 a 100 50 90 1 1 100,0" stroke="red" stroke-width="10" />
  <path d="M 740,550 A 100 50 -45 0 0 550,500" stroke="red" stroke-width="10" />
</svg>
```
[...]

PR #153, opened 2021-05-24
Support rgb colors with percentage

### LINKED ISSUE #150: RGB percentage syntax not supported
The SVG standard allows a color to be specified by the hex values e.g. `#ff00ff` as well as the RGB value `rgb(255, 0, 255)`. For the latter, percentage values are also supported e.g. `rgb(100%, 0%, 100%)`. SNV doesn't seem to support the percentage values.

```
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 1000 1000" width="1000" height="1000">
  <rect x="100" y="100" width="200" height="200" fill="rgb(255, 255, 0)" />
  <rect x="400" y="100" width="200" height="200" fill="rgb(100%, 100%, 0%)" />
</svg>
```
Rendering by Librsvg:
image
Rendering by SNV:
image

moazin:
Fixes #150. @dirkschulze Please let me know if you want me to code this differently.

### DIFF
--- a/svgnative/src/SVGStringParser.cpp
+++ b/svgnative/src/SVGStringParser.cpp
@@ -91,6 +91,24 @@ inline bool ParseDigit(CharIt& pos, const CharIt& end, std::int32_t& digit)
     return true;
 }
 
+inline bool ParseDigitOrPercentage(CharIt& pos, const CharIt& end, std::int32_t& digit, bool& percentage)
+{
+    percentage = false;
+    if (pos == end || !isDigit(*pos))
+        return false;
+    while (pos < end && isDigit(*pos))
+    {
+        digit *= 10;
+        digit += static_cast<std::int32_t>(*pos++ - '0');
+    }
+    if (pos < end && *pos == '%')
+    {
+      percentage = true;
+      pos++;
+    }
+    return true;
+}
+
 static bool ParseScientificNumber(CharIt& pos, const CharIt& end, float& number)
 {
     if (pos == end)
@@ -712,7 +730,7 @@ void ParsePathString(const std::string& pathString, Path& p)
                 currentY += newY;
             }
 
-            SVG_ASSERT(angle == 0);
+            //SVG_ASSERT(angle == 0);
 
             ArcToCurve(p, startX, startY, rx, ry, angle, flagLarge, flagSweep, currentX, currentY, prevControlX, prevControlY);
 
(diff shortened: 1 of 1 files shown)


svgnative/src/SVGStringParser.cpp
```diff
@@ -712,7 +730,7 @@ void ParsePathString(const std::string& pathString, Path& p)
                 currentY += newY;
             }
 
-            SVG_ASSERT(angle == 0);
+            //SVG_ASSERT(angle == 0);
```
dirkschulze, line 733: We should remove it entirely.

svgnative/src/SVGStringParser.cpp
```diff
@@ -936,25 +954,37 @@ static bool ParseColor(CharIt& pos, const CharIt& end, ColorImpl& paint, bool su
             std::int32_t r{};
             std::int32_t g{};
             std::int32_t b{};
+            bool r_percentage = false;
```
dirkschulze, line 957: The rest of the code uses camel case. Also, nit, use intialization lists.
```cpp
bool rPercentage{};
```

dirkschulze: Oh, another nit, boolean values should be descriptive:

`hasRPercentage` or `isRPrercentage`... depending on the context.

svgnative/src/SVGStringParser.cpp
```diff
@@ -936,25 +954,37 @@ static bool ParseColor(CharIt& pos, const CharIt& end, ColorImpl& paint, bool su
...
             if (!SkipOptWspDelimiterOptWsp(pos, end))
                 return false;
-            if (!ParseDigit(pos, end, b))
+            if (!ParseDigitOrPercentage(pos, end, b, b_percentage))
                 return false;
             if (!SkipOptWsp(pos, end))
                 return false;
             if (*pos++ != ')')
                 return false;
-            color[0] = std::min(255, r) / 255.0f;
```
dirkschulze, line 955: I'd suggest something like:

```cpp
// All values must either be digits or percentages.
if (hasRPercentage != hasGPercentage || hasRPercentage  != hasBPercentage)
    return false;
float base = hasRPercentage ? 100.0f : 255.0f;
color[0] = std::min(base, r) / base;
color[1] = std::min(base, g) / base;
color[2] = std::min(base, b) / base;
```

svgnative/src/SVGStringParser.cpp
dirkschulze:
```cpp
bool hasRPercentage{};
bool hasGPercentage{};
bool hasBPercentage{};
```

svgnative/src/SVGStringParser.cpp
moazin: Are you sure, I filed an issue #134 that shows problems with arcs, so I think the assertion is there for a good reason. I commented it out by mistake, I'm removing the comment and putting the assertion back for now.

moazin:
I think this needs some more work. I think the RGB syntax does allow for floating point values as well. So for example, `rgb(50.55%, 24.56%, 64.2%)` should be possible too right? The code I wrote only allows for integer percentages.

svgnative/src/SVGStringParser.cpp
dirkschulze: Sounds good then.

dirkschulze:
@moazin Yes, you are right, percentage values may take floats. https://drafts.csswg.org/css-color-3/#rgba-color

And yes, we should support it. 2 choices:
1. Open a new issue to support percentage with floats and only support digits for now
2. Do all in the same patch.

moazin:
@dirkschulze Done. I've tested it with a few examples, seems to work well, please review. :)

dirkschulze (approved):

PR #158, opened 2021-05-27
fix skia call to saveLayerAlpha

### LINKED ISSUE #154: Problems with opacity in Cairo and Skia ports
The support for nested `opacity` attribute has problems in the Skia and Cairo port.

Take the following document as an example.
```svg
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 2000 1000" width="2000" height="1000">
  <rect x="100" y="100" width="600" height="600" fill="peachpuff"/>
  <g opacity="0.2">
    <rect x="300" y="300" width="200" height="200"/>
  </g>
</svg>
```

Results are shown below. The reason why it doesn't work with Cairo is simply because the attributes are not nested. @dirkschulze You mentioned some functions that can handle transparency layers in Cairo. I couldn't find them. Could you please point them out? In case of Skia, the problem is simply wrong arguments to `saveLayerAlpha` in `Renderer::Save` function. The function expects an argument between 0-255 and SNV thinks it expects between 0-1.

Rendering by librsvg:
image
Rendering by SNV (Cairo):
image
Rendering by SNV (Skia):
image

moazin:
This fixes #154 partially. I've done the Skia fix. 

@dirkschulze For cairo, you pointed out that they have APIs for transparency layers. Could you point them to me, I couldn't find something simple.

### DIFF
--- a/svgnative/src/ports/skia/SkiaSVGRenderer.cpp
+++ b/svgnative/src/ports/skia/SkiaSVGRenderer.cpp
@@ -175,7 +175,7 @@ void SkiaSVGRenderer::Save(const GraphicStyle& graphicStyle)
 {
     SVG_ASSERT(mCanvas);
     if (graphicStyle.opacity != 1.0)
-        mCanvas->saveLayerAlpha(nullptr, graphicStyle.opacity);
+        mCanvas->saveLayerAlpha(nullptr, static_cast<U8CPU>(graphicStyle.opacity * 255));
     else
         mCanvas->save();
     if (graphicStyle.transform)


dirkschulze (approved):

moazin:
@dirkschulze This looks good, however, the problem still persists in Cairo. I can do the same nested `mOpacityStack` thing that I did in GDI+. You mentioned that there's an equivalent call to `saveLayerAlpha` for Cairo, but I couldn't find it. Can you point to it? If one doesn't exist, I can do the same for Cairo as I did it for GDI+.

dirkschulze:
@moazin You can use
```cpp
cairo_push_group(cr);
// put cr on a stack so you can call cairo_push_group multiple times
// to create more layers. Needed for nested opacity.
```
 to create a new layer and then draw it back with 
```cpp
cairo_pop_group_to_source(cr);
cairo_paint_with_alpha(cr, crFromTopOfStack);
```
You can take a look at the WebKit code I wrote years ago:
https://github.com/WebKit/WebKit/blob/main/Source/WebCore/platform/graphics/cairo/CairoOperations.cpp#L1198

Issue #106, opened 2020-01-20
Providing visual bouds of SVG content


dirkschulze:
For some use cases it is required to provide the visual bounds of the SVG content so that a big-enough image buffer can get created before rendering the SVG content.

Skia, CG and Cairo Graphics do support mechanisms to compute the visual bounds of the content. However, for some it might be required to traverse the entire tree while others provide a "recording" mechanism.

Cairo Graphics and Skia ([example 1](https://skia.org/user/api/SkPicture_Reference#Picture_Recorder), [API](https://github.com/google/skia/blob/master/src/core/SkPictureRecorder.cpp#L28)) do have a "recording mechanism". It means that drawing commands get "recorded" and "replayed" on another rendering context.
Both seem to have the possibility to provide the "fill bounds", the visual area.

We should consider computing at least a rough estimation of the bounds as well. The rendering port would at least (minimum requirement) need to provide the tight bounding box of paths.

dirkschulze:
@moazin @mpsuzuki any thoughts? Would recording and replaying be enough initially? Would you rather want to rely on recording for Skia and Cairo or let SVG Native Viewer compute the bounds? The latter _might_ be more expensive. But it requires meassuring.

_Note that I haven't tested `SkPicture_Recorder` yet._

moazin:
@dirkschulze A rough estimation of the bound shall suffice. The method FreeType uses for traditional glyphs also calculates a rough estimation because it's faster and almost always very close to the actual bound. 

As I understand it, relying on Cairo/Skia for the bound calculation will limit the users of FreeType to compile svg-native-viewer with only Cairo/Skia ports and the feature becomes unavailable if the user doesn't want to compile with these backends. @lemzwerg How could this affect the users of FreeType (OT-SVG)?

lemzwerg:
Sorry for being so late in replying.  I'm slowly recovering from a conference, a severe low back pain, teaching, and concerts...

I prefer if svgnative is really stand-alone with everything to properly provide all data necessary for SVG rendering, independent of the backend.  So yes, I favour a built-in recording mechanism.

moazin:
@dirkschulze Any update on this?

apodtele:
The true tight bounding box might be very hard to compute short of actual rendering. Instead, it should be rather straightforward to compute the control box which encapsulates all control points for all paths, perhaps clipped using clipPath control box as well. This is just something quick and dirty.

moazin:
[...]

PR #159, opened 2021-07-30
Adds bound calculation functionality

moazin:
## Description
This pull request attempts to add bounding box calculation functionality to SVG Native Viewer. For now, the functionality only exists in the Skia port but will be soon extended to other ports as well. In order to ensure the testing of code, I've added some unit tests as well as functionality tests that works on a comprehensive set of SVG documents and compares their bounds against the standard bounds found from using librsvg with the cairo port.

## Related Issue
#106 

## Motivation and Context
In order to render SVG glyphs properly, one needs to know the ink-box of a glyph or certain elements within it that are identified by a glyph ID.

## How Has This Been Tested?
I have unit testing for the new classes I added, namely `Rect` and the supporting `Interval`. Other than that, I have a set of SVG documents for testing purposes in the `svgnative/tests/bounds-tests-svgs` directory. Standard bounds for them are calculated by a librsvg based program that lives in the `svgnative/tests/librsvg-bounds-generator` and can write them in a file, by default I keep the files commited, thus librsvg is not a dependency unless you change the tests, in which case you'd need to run it again to update the bounds. For each file, we calculate the bounds through the Skia port and then compare it with the standard one. Some tests will fail due to existing issues in SVG Native Viewer, some that have been resolved but not yet merged other that haven't been resolved yet.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have signed the [Adobe Open Source CLA](https://opensource.adobe.com/cla.html).
- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [ ] All new and existing tests passed.


dirkschulze:
Thanks a lot @moazin ! I try to get through all changes within the next couple of days.

Could you also modify .circleci/config.yml so that we run the unit tests on every branch? Especially this branch?

moazin:
@dirkschulze Yes, I'll. This is still draft so I'm making changes to make the styling and documentation better. But meanwhile do review all the code from functionality point of view. :)

I'll modify the config, please also tell me if we can use Cmake in the tests? I'm using FetchContent to fetch GTest framework and it's not available in the Cmake version that's available in Stretch.

svgnative/src/SVGDocument.cpp
```diff
@@ -126,6 +126,22 @@ void SVGDocument::Render(const char* id, const ColorMap& colorMap, float width,
     mDocument->Render(id, colorMap, width, height);
 }
 
+Rect SVGDocument::Bounds()
+{
+    // TODO: Should we return an exception or fire an assertion for this or just return an empty rectangle?
+    if (!mDocument)
+        return Rect{};
```
moazin, line 133: @dirkschulze What do you suggest about this? In such cases, should we return an empty rectangle or throw an exception? The thing is that giving an empty rectangle is kind of misleading.

dirkschulze: Yeah, I know what you mean. Alternatively, you could pass a Rect parameter by-reference and return a boolean to indicate success.

But IMO an empty rectangle also indicates that there is an issue.

svgnative/include/svgnative/SVGDocument.h
```diff
@@ -174,6 +175,33 @@ class SVG_IMP_EXP SVGDocument
...
 
+    /**
+     * Retrieves the bounds of the SVG Document.
+     *
+     * The bounds are as tightly computed as possible. The bounds are based on the SVG canvas
+     * and don't take into account the viewBox on the root node. The rendering port's canvas
+     * is used to calculate the bounds so the user must make sure that a context has been
+     * set, otherwise an assertion will fire.
+     */
+    Rect Bounds();
```
dirkschulze, line 186: Should this be pure virtual? Is it needed to have SVGDocument::Bounds?

dirkschulze: Name suggestion GetBoundingBox(),

also, can it be const?

svgnative/include/svgnative/SVGDocument.h
```diff
@@ -174,6 +175,33 @@ class SVG_IMP_EXP SVGDocument
...
+
+    /**
+     * Retrieves the bounds of the subtree of an element with the given XML ID.
+     *
+     * The bounds are as tightly computed as possible. The bounds are based on the SVG canvas
+     * and don't take into account the viewBox on the root node. The rendering port's canvas
+     * is used to calculate the bounds so the user must make sure that a context has been
+     * set, otherwise an assertion will fire.
+     */
+    Rect Bounds(const char *id);
```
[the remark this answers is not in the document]
dirkschulze: Same suggestions and questions as for `Bounds()`

svgnative/include/svgnative/SVGRenderer.h
```diff
@@ -104,20 +104,113 @@ using Paint = boost::variant<Color, Gradient>;
...
+ * 4. The join of two intervals $A$ and $B$ is defined as an interval whose
+ * set is defined as the smallest possible set of *complete* real numbers that
+ * contain all members of $A$ as well as $B$. By complete we simply mean that
+ * for any possible two members of the set, all real numbers between the two
+ * must also be within the set. The word smallest here means that no other proper
+ * subset of that set will satisfy all the requirements.
+ * 5. The intersection of two intervals $A$ and $B$ is the one whose set is
+ * the intersection of the sets of $A$ and $B$.
+ */
+class Interval
```
dirkschulze, line 128: Didn't get to review how Interval is used yet, but it seems that it could be a private header and doesn't need to be in public headers.

svgnative/include/svgnative/SVGRenderer.h
```diff
@@ -104,20 +104,113 @@ using Paint = boost::variant<Color, Gradient>;
...
+     * contains both the intervals */
+    Interval operator|(Interval other);
+
+    /* Returns true if one interval contains another one */
+    bool Contains(Interval other);
+
+    /* Is the interval empty? having no points in its set? */
+    bool IsEmpty() { return a == b; }
+
+    float a = 0;
```
dirkschulze, line 161: should be private.

svgnative/include/svgnative/SVGRenderer.h
```diff
@@ -104,20 +104,113 @@ using Paint = boost::variant<Color, Gradient>;
...
+
+    /* Computes the intersection of this interval with another one */
+    Interval operator&(Interval other);
+
+    /* Computes the join of this interval with another one, think of an interval that
+     * contains both the intervals */
+    Interval operator|(Interval other);
+
+    /* Returns true if one interval contains another one */
+    bool Contains(Interval other);
```
dirkschulze, line 156: const?

svgnative/include/svgnative/SVGRenderer.h
```diff
@@ -104,20 +104,113 @@ using Paint = boost::variant<Color, Gradient>;
...
+
+    /* Is the interval empty? having no points in its set? */
+    bool IsEmpty() { return a == b; }
```
dirkschulze, line 159: const

svgnative/include/svgnative/SVGRenderer.h
```diff
@@ -104,20 +104,113 @@ using Paint = boost::variant<Color, Gradient>;
...
-        : x{aX}
-        , y{aY}
-        , width{aWidth}
-        , height{aHeight}
-    {
-    }
-    float x = std::numeric_limits<float>::quiet_NaN();
-    float y = std::numeric_limits<float>::quiet_NaN();
-    float width = std::numeric_limits<float>::quiet_NaN();
-    float height = std::numeric_limits<float>::quiet_NaN();
```
dirkschulze, line 120: We used these values on purpose to differ between actual empty rects and CTOR created rects, if you need {0, 0, 0, 0} I'd suggest to explicitly call the CTOR with 0s

svgnative/include/svgnative/SVGRenderer.h
```diff
@@ -104,20 +104,113 @@ using Paint = boost::variant<Color, Gradient>;
...
+    Rect(float aX, float aY, float aWidth, float aHeight);
+    /* Returns if the rectangle is empty */
+    bool IsEmpty();
```
dirkschulze, line 192: const

svgnative/include/svgnative/SVGRenderer.h
```diff
@@ -104,20 +104,113 @@ using Paint = boost::variant<Color, Gradient>;
...
+    /* Returns the two intervals defining the rectangle */
+    IntervalPair Intervals() { return IntervalPair(Interval(x, x + width), Interval(y, y + height)); }
+    /* Returns true if Rect contains the other Rect within it */
+    bool Contains(Rect other);
```
dirkschulze, line 196: const

svgnative/include/svgnative/SVGRenderer.h
```diff
@@ -104,20 +104,113 @@ using Paint = boost::variant<Color, Gradient>;
...
-    }
-    float x = std::numeric_limits<float>::quiet_NaN();
-    float y = std::numeric_limits<float>::quiet_NaN();
-    float width = std::numeric_limits<float>::quiet_NaN();
-    float height = std::numeric_limits<float>::quiet_NaN();
+    Rect(float aX, float aY, float aWidth, float aHeight);
+    /* Returns if the rectangle is empty */
+    bool IsEmpty();
+    /* Returns the two intervals defining the rectangle */
+    IntervalPair Intervals() { return IntervalPair(Interval(x, x + width), Interval(y, y + height)); }
```
dirkschulze, line 194: const

svgnative/include/svgnative/SVGRenderer.h
dirkschulze: For most other functions as well

svgnative/src/SVGDocumentImpl.cpp
```diff
@@ -973,6 +973,128 @@ void SVGDocumentImpl::RenderElement(const Element& element, const ColorMap& colo
     SVG_ASSERT(mVisitedElements.empty());
 }
 
+Rect SVGDocumentImpl::Bounds()
+{
+    SVG_ASSERT(mGroup);
+    // TODO: Should we fire an assertion or raise exception?
```
dirkschulze, line 979: IIRC `SVGDocument` can only be called with an actual SVG data. So not sure if either is needed.

svgnative/src/SVGDocumentImpl.cpp
```diff
@@ -973,6 +973,128 @@ void SVGDocumentImpl::RenderElement(const Element& element, const ColorMap& colo
...
+    if (!mGroup)
+        return Rect{};
+    ExtractBounds(*mGroup);
+    Rect sum_bound;
```
dirkschulze, line 983: camelCase

svgnative/src/SVGDocumentImpl.cpp
```diff
@@ -973,6 +973,128 @@ void SVGDocumentImpl::RenderElement(const Element& element, const ColorMap& colo
...
+    for(auto const &bound: mBounds)
```
dirkschulze, line 984:
```cpp
for (auto const& bound : mBounds)
```

Don't we have styling enabled for testing?

svgnative/src/SVGDocumentImpl.cpp
```diff
@@ -973,6 +973,128 @@ void SVGDocumentImpl::RenderElement(const Element& element, const ColorMap& colo
...
+}
+
+Rect SVGDocumentImpl::Bounds(const char* id)
+{
+    SVG_ASSERT(mGroup);
+    // TODO: Should we fire an assertion or raise exception?
+    if (!mGroup)
+        return Rect{};
+
+    auto elementIter = mIdToElementMap.find(id);
```
dirkschulze, line 998: const auto

svgnative/src/SVGDocumentImpl.cpp
```diff
@@ -973,6 +973,128 @@ void SVGDocumentImpl::RenderElement(const Element& element, const ColorMap& colo
...
+    if (elementIter != mIdToElementMap.end())
+        ExtractBounds(*elementIter->second);
+    else
+    {
+        // TODO: if such an element does not exist, should we raise an exception?
```
dirkschulze, line 1003: Assertion at least

svgnative/src/SVGDocumentImpl.h
```diff
@@ -240,6 +247,7 @@ class SVGDocumentImpl
     // hierarchy.
     std::stack<StrokeStyleImpl> mStrokeStyleStack;
     std::stack<FillStyleImpl> mFillStyleStack;
+    std::vector<Rect> mBounds;
```
dirkschulze, line 250: I wonder if we really need to collect all bounds or if it wouldn't be better to compute union/intersection on the fly.

A document might be huge with many elements, better to free as many Rects as possible while traversing the tree.

svgnative/src/SVGRenderer.cpp
```diff
@@ -0,0 +1,107 @@
+#include "svgnative/SVGRenderer.h"
+#include <tuple>
+#include <cmath>
+#include <stdexcept>
+
+namespace SVGNative {
+    Interval::Interval(float u, float v)
```
dirkschulze, line 7: Interval logic should be in separate files

svgnative/src/ports/skia/SkiaSVGRenderer.cpp
```diff
@@ -67,7 +67,7 @@ void SkiaSVGPath::CurveTo(float x1, float y1, float x2, float y2, float x3, floa
 
 void SkiaSVGPath::CurveToV(float x2, float y2, float x3, float y3)
 {
-    mPath.cubicTo(mCurrentX, mCurrentY, x2, y2, x3, y3);
+    mPath.quadTo(x2, y2, x3, y3);
```
dirkschulze, line 70: Would be better to raise a different bug for it.

svgnative/include/svgnative/SVGRenderer.h
dirkschulze: Ah ok, an Interval seems to be a coordinate pair.

Maybe, instead of `x`, `y`, `width`, `height` we should use `x1`, `y1` and `x2`, `y2` on `Rect` directly.

Ideally, all the logic would then be hidden and the changes to the Rect class are as minimal as possible.

dirkschulze (commented):
IMO this looks very good, in general I'd suggest that we do not collect all rectangles but instead compute them on the fly. The earlier we free the Rects we need to create, the better.

moazin:
@dirkschulze Thank you for all your suggestions, I'll go through them soon, I've been cleaning up an MR on FreeType side as well.

Could you please also take a look at the testing code I wrote?

svgnative/src/SVGDocumentImpl.h
moazin: I agree that this is more compute intensive. The only reason why I didn't do them on the fly is because I have a debugging tool built as an example of SVG Native Viewer that consumes another API function that returns all of these rectangles and it becomes very useful to see which rectangle is messing up the bounds and where it comes from. Otherwise, one has to manually debug in GDB to figure out which shape is doing what. Of course that's a testing only thing so I can switch to doing this on the fly, saving all of these does cause extra memory.

dirkschulze: Yeah, unless it is too much code, you may put the debugging logic behind a debug preprocessor flag.

svgnative/include/svgnative/SVGDocument.h
moazin: By virtual you mean defined in header-only?

About the name, the rest of the functions such as Width and Height don't have a Get? So does it make sense to name it GetBoundingBox? It is for sure more meaningful so I don't have any problem changing it.

moazin: Also, is an empty rectangle really indicative of a problem? One can have dumb paths like `<path d="M 100,100" />` or the standard doesn't allow that?

svgnative/include/svgnative/SVGRenderer.h
moazin: @dirkschulze Umm, so interval is basically a supporting class that makes life easier for rectangle intersections and joins. It doesn't need to be used by the client applications, `Rect` is public however and  it has a function that gets the intervals. This is used internally so maybe I can make it private.
> Maybe, instead of x, y, width, height we should use x1, y1 and x2, y2 on Rect directly.
Ideally, all the logic would then be hidden and the changes to the Rect class are as minimal as possible.

Could you elaborate this? If we simply replace width and height by x2 and y2, we have to compute width and height in function calls for those place where they are needed. That avoids computation for sure, but how does it avoid changes to the `Rect` class?

svgnative/src/SVGDocumentImpl.cpp
moazin: Styling as in CSS style or code style checks? We don't have a clang-tidy file or something like that, do we?

svgnative/include/svgnative/SVGRenderer.h
moazin: I'm doing that right now. But there can be serious problem with this. All of the rest of my functions such as `isEmpty`, `join`, `intersection` become buggy when it's an actual empty rect (as opposed to CTOR created one) as quiet_NaN doesn't work with equality or other operations I think. The programmer will have to exercise care too make sure never to create a `Rect` by doing `Rect{}`. Btw, why do you want to distinguish between the two? Or when would you create a `Rect` by doing `Rect{}`?

svgnative/src/SVGDocumentImpl.cpp
moazin: I don't think our pipeline has any code styling checks, does it? In my case it's always failing because of the CMake problem I mentioned. I am using FetchContent which is not available in stretch. It needs buster.

moazin:
What did you exactly mean? Which unit tests do you mean? The ones I wrote? Which are basically three files in `tests` directory? Or do you mean your python based testing script? I guess that should automatically run on this branch with CI. In my case tests are not progressing because of the FetchContent problem.

> Could you also modify .circleci/config.yml so that we run the unit tests on every branch? Especially this branch?

svgnative/include/svgnative/SVGDocument.h
```diff
@@ -174,6 +175,34 @@ class SVG_IMP_EXP SVGDocument
...
+     *
+     * The bounds are as tightly computed as possible. The rendering port's canvas
+     * is used to calculate the bounds so the user must make sure that a context has been
+     * set, otherwise an assertion will fire.
+     */
+    bool GetBoundingBox(const char* id, Rect& bounds);
+
+#ifdef DEBUG_API
+    /**
+     * TODO: Remove this.
```
dirkschulze, line 198: TODO ? Do we want to keep it? If it is not for the public, move it to SVGDocumentImpl.h?

svgnative/include/svgnative/SVGRenderer.h
```diff
@@ -98,26 +98,65 @@ struct Gradient;
...
+class Interval;
 
 using Color = std::array<float, 4>;
 using Paint = boost::variant<Color, Gradient>;
 using ColorStop = std::pair<float, Color>;
 using ColorMap = std::map<std::string, Color>;
+using IntervalPair = std::tuple<Interval, Interval>;
 
-struct Rect
+/**
```
dirkschulze, line 109: Could you move this out into its own header?

svgnative/src/Interval.h
```diff
@@ -0,0 +1,81 @@
...
+     * contains both the intervals */
+    Interval operator| (Interval other) const;
+
+    /* Returns true if one interval contains another one */
+    bool Contains (Interval other) const;
+
+    /* Is the interval empty? having no points in its set? */
+    bool IsEmpty() const { return a == b; }
+  private:
+    float a = 0;
```
dirkschulze, line 73: for consistency, init lists:
```cpp
float mA{ };
float mB{ };
```

svgnative/src/SVGDocumentImpl.cpp
```diff
@@ -973,6 +973,179 @@ void SVGDocumentImpl::RenderElement(const Element& element, const ColorMap& colo
...
+    GraphicStyleImpl graphicStyle{};
+    graphicStyle.transform = mRenderer->CreateTransform();
+    graphicStyle.transform->Translate(-1 * mViewBox[0], -1 * mViewBox[1]);
+    auto saveRestore = SaveRestoreHelper{mRenderer, graphicStyle};
+    ExtractBounds(*mGroup);
+    SVG_ASSERT(mVisitedElements.empty());
+
+    Rect sumBound{0, 0, 0, 0};
+#ifdef DEBUG_API
+    for(auto const& bound : mBounds)
```
dirkschulze, line 991:
```cpp
for (auto const& bound : mBounds)
```

svgnative/include/svgnative/SVGRenderer.h
```diff
@@ -98,26 +98,65 @@ struct Gradient;
 class Transform;
 class Path;
 class Shape;
+class Interval;
```
dirkschulze, line 101: Can be moved to the new Rect.h header

svgnative/include/svgnative/SVGRenderer.h
```diff
@@ -98,26 +98,65 @@ struct Gradient;
...
 
 using Color = std::array<float, 4>;
 using Paint = boost::variant<Color, Gradient>;
 using ColorStop = std::pair<float, Color>;
 using ColorMap = std::map<std::string, Color>;
+using IntervalPair = std::tuple<Interval, Interval>;
```
dirkschulze, line 107: Can be moved to the new header.

svgnative/include/svgnative/SVGRenderer.h
```diff
@@ -98,26 +98,65 @@ struct Gradient;
...
+    float Top() const { return y; }
+    float Bottom() const { return y + height; }
+
     float x = std::numeric_limits<float>::quiet_NaN();
     float y = std::numeric_limits<float>::quiet_NaN();
     float width = std::numeric_limits<float>::quiet_NaN();
     float height = std::numeric_limits<float>::quiet_NaN();
+
+  private:
+    IntervalPair Intervals() const;
```
dirkschulze, line 159: Do we need to store this in the Rect class? Ideally, Interval would not be visible to any client. Maybe it could be inlined in a private header?

svgnative/src/SVGRenderer.cpp
```diff
@@ -0,0 +1,66 @@
+#include "svgnative/SVGRenderer.h"
+#include "Interval.h"
+#include <tuple>
+#include <cmath>
+#include <stdexcept>
+
+namespace SVGNative
+{
+    Rect::Rect(float aX, float aY, float aWidth, float aHeight)
```
dirkschulze, line 9: This should move into Rect.cpp

svgnative/src/SVGRenderer.cpp
```diff
@@ -0,0 +1,66 @@
...
+    {
+        if (aWidth < 0 || aHeight < 0)
+        {
+            throw std::invalid_argument("Height or Width of a rectangle can't be negative!");
+        }
+        x = aX;
+        y = aY;
+        width = aWidth;
+        height = aHeight;
```
dirkschulze, line 18: We should use mX, mY, mWidth and mHeight for the members.

svgnative/src/SVGRenderer.cpp
```diff
@@ -0,0 +1,66 @@
...
+        return std::get<0>(this_intervals).Contains(std::get<0>(other_intervals)) &&
+            std::get<1>(this_intervals).Contains(std::get<1>(other_intervals));
+    }
+    bool Rect::operator==(Rect other) const
+    {
+        return x == other.x && y == other.y && width == other.width && height == other.height;
+    }
+    Rect Rect::operator&(Rect other) const
+    {
+        IntervalPair intervals_a = Intervals();
```
dirkschulze, line 38: CamelCase names

svgnative/src/SVGDocumentImpl.cpp
tjindal: @dirkschulze removed DEBUG_API contents.

svgnative/src/Interval.h
tjindal: @dirkschulze above format of init is used

svgnative/include/svgnative/SVGDocument.h
tjindal: @dirkschulze removing  DEBUG_API content from this file

svgnative/include/svgnative/SVGRenderer.h
tjindal: a new file rect.h and rect. cpp is made and moved the relevant contents.

svgnative/src/SVGRenderer.cpp
tjindal: in rect.cpp it is moved

svgnative/src/SVGRenderer.cpp
tjindal: Yes, we need to use like mX ,mY but due to public nature of these variables, these are used many where in code so if we do this change, we need to change in multiple files and functions.For now, i kept as it is .

svgnative/src/SVGRenderer.cpp
tjindal: camelCase names done

PR #165, opened 2022-04-18
Include <string> in ExpatXMLParser.cpp

bungeman:
Both `<string>` and `<string.h>` are needed in ExpatXMLParser.cpp.


dirkschulze:
LGTM. Thanks @bungeman

PR #166, opened 2022-04-18
Explicitly define ~SkiaSVGImageData

bungeman:
The `SkiaSVGImageData` destructor must destroy an `sk_sp<SkImage>` member.
However, `SkiaSVGRenderer.h` only has a forward declaration of `SkImage` so
the default inline destructor cannot destruct the `SkImage` unless the
translation unit which includes `SkiaSVGRenderer.h` pulls in the full
definition of `SkImage`. Remove this awkward situation by defining the
destructor in `SkiaSVGRenderer.cpp`, which does have a full definition of
`SkImage`.


dirkschulze:
LGTM and thanks for the detailed description of the issue! Thanks @bungeman

PR #167, opened 2022-04-18
Include SkEncodedImageFormat.h in TestSkia.cpp.

bungeman:
Some IWYU to allow TestSkia.cpp to build with newer versions of Skia
where SkEncodedImageFormat is not implicitly included through other
includes.


dirkschulze:
LGTM. Thanks for the PR!

PR #168, opened 2022-04-18
Check SkCodec was created before getting origin.

bungeman:
If the image data passed to SkCodec::MakeFromData is not recognized as
image data the SkCodec::MakeFromData may return nullptr. Check for this
failure and return instead of crashing.


dirkschulze:
Thanks for the patch @bungeman . LGTM.

PR #176, opened 2022-06-07
support number or percentage values for offset/opacity values 

tjindal:
Support of Percentage values when specifies offset/opacity for stroke/fill properties 
Make a ParseNumber function generic, now work for both as a number or % values
Modified testcase which initially does not support % values

### DIFF
--- a/svgnative/src/SVGDocumentImpl.cpp
+++ b/svgnative/src/SVGDocumentImpl.cpp
@@ -611,7 +611,7 @@ void SVGDocumentImpl::ParseFillProperties(FillStyleImpl& fillStyle, const Proper
     if (prop != iterEnd)
     {
         float opacity{};
-        if (SVGStringParser::ParseNumber(prop->second, opacity))
+        if (SVGStringParser::ParseNumberOrPercentage(prop->second, opacity))
             fillStyle.fillOpacity = std::max<float>(0.0, std::min<float>(1.0, opacity));
     }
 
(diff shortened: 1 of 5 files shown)


svgnative/src/SVGStringParser.cpp
```diff
@@ -400,7 +400,7 @@ bool ParseLengthOrPercentage(const std::string& lengthString, float relDimension
     return !SkipOptWsp(pos, end);
 }
 
-bool ParseNumber(const std::string& numberString, float& number)
+bool ParseNumberOrPercentage(const std::string& numberString, float& number)
```
dirkschulze, line 403: Please keep ParseNumber. Also, I am confused it is not used in more places.

svgnative/src/SVGStringParser.cpp
```diff
@@ -409,6 +409,11 @@ bool ParseNumber(const std::string& numberString, float& number)
         return false;
     if (!ParseScientificNumber(pos, end, number))
         return false;
+    if (pos < end && *pos == '%')
+    {
+        number = number/100;
```
dirkschulze, line 414: This works for opacity where % is actually based on 1.f. But there could be other uses. I'd suggest to have a function `ParseAlphaValue()` so that it is clear why this is ok.

Also, do we use double or single precision? If the former, please add `.f` otherwise `.` at the end.

dirkschulze (changes requested):

svgnative/src/SVGDocumentImpl.cpp
```diff
@@ -703,7 +703,7 @@ void SVGDocumentImpl::ParseStrokeProperties(StrokeStyleImpl& strokeStyle, const
     {
         float miter{};
         // Miter must be bigger 1. Otherwise ignore.
-        if (SVGStringParser::ParseNumber(prop->second, miter) && miter >= 1)
+        if (SVGStringParser::ParseNumberOrPercentage(prop->second, miter) && miter >= 1)
```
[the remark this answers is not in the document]
tjindal: Yes i completely agree, there are no percentages for stroke-miterlimit. as i make a function generic which will works for both when you provide a number or we provide a percentage . so in this case , it will simply ignore the newly lines that added as when we have number only, in that case (pos == end) both are same.

dirkschulze (approved):

Issue #205, opened 2025-08-26
Exception raised when string parser dereferences end iterator


thinlang:
In certain cases in newer photoshop code, I noticed the string parser will step its string iterator to end or beyond and then dereference it, which subsequently raises an exception. May be the same issue as https://github.com/adobe/svg-native-viewer/issues/204.

PR #206, opened 2025-08-26
Fix for issue 205: the string parser code overstepping to (or past) "end()"

thinlang:
## Description

When parsing strings, the code sometimes oversteps the string iterator to end or past end and then dereferences without checking if it's valid to do so. This results in an exception being thrown in newer versions of Visual Studio.

The change involves modifying functions that take a current position and end iterator. Instead, they're passed an object that tracks both of those and ensures that:
1. The iterator never steps past end
2. The iterator is only dereferenced when it is safe to do so

This change also includes some cleanup of warnings on Windows

## Related Issue

https://github.com/adobe/svg-native-viewer/issues/205

## Motivation and Context

String parsing should not throw exceptions for reasonable or badly formed files.

## How Has This Been Tested?

I ran the test suite locally on Windows per the readme instructions.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have signed the [Adobe Open Source CLA](https://opensource.adobe.com/cla.html).
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.


svgnative/CMakeLists.txt
```diff
@@ -124,7 +124,7 @@ elseif(MSVC)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /we4302 /we4189")
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /wd4290 /wd4335 /wd4355 /wd4814 /wd4091 /TP")
     if (NOT USE_SKIA)
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /std:c++11")
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /std:c++14")
```
dirkschulze, line 127: The idea is to have an interoperable library that is able to go back to C++11. So a requirement for C++14 should be avoided.

We do have code that allows to switch to C++17 but keeping C++11 compatibility. See PR #194.

svgnative/src/SVGParserArcToCurve.cpp
```diff
@@ -106,7 +106,7 @@ void ArcToCurve(SVGNative::Path& path, float startX, float startY, float radiusX
         return;
     }
 
-    SVG_ASSERT(radiusX == radiusY); // untested
+    //SVG_ASSERT(radiusX == radiusY); // untested
```
dirkschulze, line 109: Why remove this assertion? Float comparison should include epsilon, but in general?

svgnative/src/SVGStringParser.cpp
```diff
@@ -29,54 +29,139 @@ void ArcToCurve(Path& path, float startX, float startY, float radiusX, float rad
...
+    auto operator+(T) = delete;
 
-inline bool isHex(char c)
-{
-    return isDigit(c) || (c >= 'a' && c <= 'f') ||  (c >= 'A' && c <= 'F');
-}
+    template <typename T>
+    auto operator-(T) = delete;
+
+    [[nodiscard]] explicit operator bool() const
```
dirkschulze, line 46: This is not compatible to C++11. If you think we should add it regardless, think about compatible ways like with such macro setup:

```cpp
#if defined(__has_cpp_attribute)
#  if __has_cpp_attribute(nodiscard) && __cplusplus >= 201703L
#    define NODISCARD [[nodiscard]]
#  elif defined(__GNUC__) || defined(__clang__)
#    define NODISCARD __attribute__((warn_unused_result))
#  elif defined(_MSC_VER)
#    define NODISCARD _Check_return_
#  else
#    define NODISCARD
#  endif
#else
#  if defined(__GNUC__) || defined(__clang__)
#    define NODISCARD __attribute__((warn_unused_result))
#  elif defined(_MSC_VER)
#    define NODISCARD _Check_return_
#  else
#    define NODISCARD
#  endif
#endif
```

svgnative/src/SVGStringParser.cpp
```diff
@@ -29,54 +29,139 @@ void ArcToCurve(Path& path, float startX, float startY, float radiusX, float rad
...
+            _c = *_pos;
+        }
+    }
 
-inline bool isDigit(char c)
-{
-    return c >= '0' && c <= '9';
-}
+    template <typename T>
+    auto operator+(T) = delete;
```
dirkschulze, line 41: This is C++20, no? So not valid in C++11 either.

svgnative/src/SVGStringParser.cpp
```diff
@@ -29,54 +29,139 @@ void ArcToCurve(Path& path, float startX, float startY, float radiusX, float rad
 
 namespace SVGStringParser
 {
-using CharIt = std::string::const_iterator;
+class CharPos final {
```
dirkschulze, line 32: While I do like avoiding the `endPos` check, this can lead to a bigger performance degradation in a very performance critical path. We should run performance tests with huge path strings.

svgnative/src/SVGParserArcToCurve.cpp
thinlang: This assert fired on my machine when running the tests on Windows locally

